PR #4, opened 2024-06-18
Use hatch-vcs for versioning pyodide-build

ryanking13:
Instead of updating the version number manually, this use `hatch-vcs` and use git metadata to handle the version. Just same as micropip and pytest-pyodide.


hoodmane (approved):
Looks fine to me. I don't really know how this hatch stuff works.

PR #8, opened 2024-07-02
Allow overriding _f2c_fixes.py from user

ryanking13:
While most parts of pyodide-build are decoupled with pyodide/pyodide, `_f2c_fixes.py` file that is used to build scipy is highly coupled with the scipy recipe in pyodide/pyodide.

It would be very inefficient if we need to update pyodide-build whenever we fix scipy (e.g. https://github.com/pyodide/pyodide/issues/4818). Therefore, this add a private config variable `_f2c_fixes_wrapper`, which allows users to modify override `_f2c_fixes.py` file without touching pyodide-build itself.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,7 +16,7 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: "v0.3.5"
+    rev: "v0.5.0"
     hooks:
       - id: ruff
         args: [--fix]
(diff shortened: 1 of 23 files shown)


hoodmane:
Interesting, thanks @ryanking13. Looks reasonable for now. I'm currently working on making a fork of f2c with modifications that suit our needs so hopefully we can get rid of f2c_fixes.

ryanking13:
> I'm currently working on making a fork of f2c with modifications that suit our needs so hopefully we can get rid of f2c_fixes.

Cool. That sound ambitious :)

> Looks reasonable for now

Yeah, for now, I am trying to remove all annoying parts when unvendoring pyodide-build.

PR #14, opened 2024-07-22
Access `sysconfig` paths through `pyodide config` CLI

agriyakhetarpal:
## Description

This PR is a re-do of https://github.com/pyodide/pyodide/pull/4947, which aimed to add the include directory containing the Python interpreter headers. This PR goes one step further and adds a way to access all of the paths listed under `sysconfig.get_paths()`. These paths are available in the default configuration values under `pyodide config list` and are accessible through `pyodide config get <variable name>`, say `pyodide config get pyodide_sysconfig_include`. These contain the `pyodide_sysconfig` prefix to prevent any potential conflicts with any of the other computed variables.

## Checklist

- [x] Tested via `TestConfigManager_OutOfTree::test_default_config`, so an additional test isn't needed
- [ ] CHANGELOG entry


agriyakhetarpal:
Personally, I'm not a fan of the `pyodide_sysconfig` prefix, but I think it helps avoid us being ambiguous. Also, I think `python_sysconfig` would be a better prefix instead of `pyodide_sysconfig` – this is ready for review, @ryanking13 (I don't have permissions to request reviews on this repository)

pyodide_build/config.py
```diff
@@ -29,8 +30,15 @@ def __init__(self, pyodide_root: Path):
...
             **self._load_config_file(Path.cwd(), os.environ),
             **self._load_config_from_env(os.environ),
+            **self._load_sysconfig_paths(),
         }
 
+    def _load_sysconfig_paths(self) -> Mapping[str, str]:
+        # Load the values of sysconfig.get_paths() with a
+        # "pyodide_sysconfig_" prefix to differentiate them
+        # from other configuration variables
+        return {f"pyodide_sysconfig_{k}": v for k, v in sysconfig.get_paths().items()}
```
ryanking13, line 40: This would retrieve the host Python's sysconfig values, not those in the cross build env. We should parse these values from the `_sysconfigdata__emscripten_wasm32-emscripten.py` file in the xbuildenv.

pyodide_build/cli/config.py
```diff
@@ -19,6 +21,13 @@
     "meson_cross_file": "MESON_CROSS_FILE",
 }
 
+# Load the values of sysconfig.get_paths() with a
+# "pyodide_sysconfig_" prefix to differentiate them
+# from other configuration variables
+PYODIDE_CONFIGS.update(
+    {f"pyodide_sysconfig_{k}": v for k, v in sysconfig.get_paths().items()}
```
ryanking13, line 28: There are only eight values in `sysconfig.get_paths()`, so, I think it is okay to list all of them, which will be more explicit and make the code more readable.

ryanking13 (commented):
Thanks @agriyakhetarpal!

> pyodide_sysconfig_*

Maybe just `sysconfig_*`? pyodide prefix seems redundant to me.

pyodide_build/config.py
agriyakhetarpal: Where should I retrieve this file from the cross-build environment? I have noted that I can get the file when I do `pyodide xbuildenv install`, but if there's already a programmatic way to access this, I can use that instead of writing my own methods.

ryanking13: Here is the code we locate the sysconfig file. Since it uses the variables in the Makefile.envs, you'll need to modify it a bit to use it in the config manager to prevent cyclic import.

https://github.com/pyodide/pyodide-build/blob/fac0109aa2acf14469320b049d710dd42639bf94/pyodide_build/pypabuild.py#L98-L102

I think you'll also need to read the [sysconfig.py](https://github.com/maxking/cpython-1/blob/master/Lib/sysconfig.py) and how it retrieves the value from the sysconfig file, as we cannot directly use `sysconfig` module here.

PR #18, opened 2024-08-12
Remove `shared_library` key

ryanking13:
This key is removed in

- pyodide: https://github.com/pyodide/pyodide/pull/4996
- pyodide-lock: https://github.com/pyodide/pyodide-lock/pull/31

and this is the last PR to remove it in pyodide-build as well.


hoodmane (approved):
Thanks!

PR #19, opened 2024-08-14
Infer python version from Makefile.envs in PyO3

ryanking13:
Instead of hardcoding the Python version, use `PYO3_CROSS_PYTHON_VERSION` env variable, which [PyO3 uses to find the correct Python include/lib paths](https://pyo3.rs/main/building-and-distribution#cross-compiling).

Also removed pyo3_config.ini file. Otherwise, the config file always take precedence.

### DIFF
--- a/pyodide_build/config.py
+++ b/pyodide_build/config.py
@@ -162,7 +162,6 @@ BUILD_KEY_TO_VAR: dict[str, str] = {
     "sysconfig_name": "SYSCONFIG_NAME",
     "targetinstalldir": "TARGETINSTALLDIR",
     "cmake_toolchain_file": "CMAKE_TOOLCHAIN_FILE",
-    "pyo3_config_file": "PYO3_CONFIG_FILE",
     "meson_cross_file": "MESON_CROSS_FILE",
     "cflags_base": "CFLAGS_BASE",
     "cxxflags_base": "CXXFLAGS_BASE",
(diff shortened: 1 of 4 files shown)


ryanking13:
It isn't working, but at least I found that integration tests are useful enough to detect this kind or errors :) Let's see if I can fix it.

ryanking13:
It is working now. I also checked it in pyodide/pyodide in my fork https://github.com/ryanking13/pyodide/commit/be7e588f6cf33fe7496e09b543b4a89753a469ef

pyodide_build/config.py
```diff
@@ -213,8 +212,9 @@ def to_env(self) -> dict[str, str]:
     "cxxflags": "$(CXXFLAGS_BASE)",
     "ldflags": "$(LDFLAGS_BASE) -s SIDE_MODULE=1",
     # Rust-specific configuration
-    "pyo3_cross_lib_dir": "$(CPYTHONINSTALL)/lib",
+    "pyo3_cross_lib_dir": "$(CPYTHONINSTALL)/sysconfigdata",  # FIXME: pyodide xbuildenv stores sysconfigdata here
```
hoodmane, line 215: What about `$(CPYTHONINSTALL)/lib/python3.12/`?

ryanking13: We don't have that directory in our xbuildenv now. Normally sysconfigdata would be placed in that directory but we don't vendor `lib/python3.X.Y/...` part in the xbuildenv and only store sysconfigdata in the top-level directory.

```
xbuildenv/pyodide-root$ ls -R | grep ":$" | sed -e 's/:$//' -e 's/[^-][^\/]*\//--/g' -e 's/^/ /' -e 's/-/|/'
 .
 |-cpython
 |---installs
 |-----python-3.12.1 (this is CPYTHONINSTALL)
 |-------include
 |---------python3.12
 |-----------cpython
 |-----------internal
 |-------sysconfigdata
 |---------__pycache__
```


Maybe we can change the path in the xbuildenv, but for now the sysconfigdata directory works like a directory where first-party libraries are stored.

ryanking13:
Thanks for the review!

PR #21, opened 2024-08-21
Install unisolated packages into the isolated build venv and overlay their cross-build files

ryanking13:
This changes how we handle unisolated packages and the hostsitepackages directory.

The current mechanism of unisolated packages works as follows:

1. We have a small number of unisolated packages. NumPy and SciPy are the most important ones.
2. These unisolated packages are often used as build dependencies for other packages.
3. When a package lists one of these as a build dependency, we intercept the call that installs build dependencies and exclude that package from being installed.
4. Instead, we point PYTHONPATH at the WASM build of the unisolated package, so it gets used in place of the native version when building.

This mechanism caused a few issues, such as the one discussed in https://github.com/pyodide/pyodide/pull/5012#discussion_r1720979108. In particular, build time scripts that these packages ship, like `f2py` and `numpy-config` from NumPy, never actually made it into the isolated build environment, so they weren't available on PATH during the build.

This PR changes it in the following way:

1. Do not exclude unisolated packages when installing build dependencies.
2. Instead, install the pinned cross-build versions of the packages like any other dependency, then replace only its cross-build files (headers, static libraries, `.pxd` files) with the WASM-compatible versions by grabbing them from the cross-build environment.

We keep the existing mechanism that installs cross-build packages into `HOSTSITEPACKAGES`. `PYTHONPATH` is no longer pointed at it, though, because Python-level access during the build (such as importing `numpy`, calling `numpy.get_include()`, running `f2py` or `numpy-config`, and so on) now comes straight from the isolated venv. 

Some more info on why we kept `HOSTSITEPACKAGES` around is noted below, in the "Testing" section.

**[EDIT: reverted this one, will split it up into its own PR]:** Another change is that `oldest-supported-numpy` is no longer silently ignored when it appears as a build dependency and instead raises an error. It has been deprecated since NumPy 2.0, so packages that still depend on it should move to a plain `numpy` dependency. If this causes trouble for a specific recipe, it can still be ignored by setting `ignored_build_requirements` to "patchelf oldest-supported-numpy"` of course, but IMO we don't need to bother about it anymore. In https://github.com/pyodide/pyodide-recipes/pull/591, only Fiona was using it. I updated its version to one that no longer uses it, and things are working fine there.

Regarding the `PKG_CONFIG_LIBDIR` change to find `numpy.pc`: this is a new issue, as SciPy 1.17.1 used to not require `numpy` explicitly (https://github.com/scipy/scipy/blob/527eb7fd7953a1de068f94bf8b322f249b9405ae/scipy/meson.build#L37), but when I merged the PR for SciPy 1.18.0 (https://github.com/pyodide/pyodide-recipes/pull/586), it has made `numpy` a hard requirement: https://github.com/scipy/scipy/blob/54ef5423f2e4376230ec3bfda6912a07a50958e3/scipy/meson.build#L34 (and https://github.com/scipy/scipy/pull/24119#issuecomment-3631045964). We use this environment variable to scan for NumPy's pkg-config file and add it to the build. `PKG_CONFIG_LIBDIR` is not additive as such, but we respect it if it's already set by a package. My reason for doing so is noted here: https://github.com/scipy/scipy/pull/24119#discussion_r2614925746.

This also adds a `numpy-scripts-example` integration test recipe. I wrote a small extension module built with meson-python that relies on `numpy-config` and `f2py` being on PATH to exercise the new mechanism. We don't compile anything Fortran-related.

## Testing

This was tested in https://github.com/pyodide/pyodide-recipes/pull/591, where all 325 package builds and tests passed.

The first run turned up an issue with building RobotRaconteur, and opencv-python would have had the same issue. Their builds are CMake-based, and they need direct access to NumPy's `ndarrayobject.h` and other headers in `numpy/_core/include/numpy/`, which used to work because `PYTHONPATH` pointed to a WASM-patched NumPy install. With that mechanism gone, those headers were nowhere to be found, and the builds failed.

A tempting fix for us would have been to add the entire `numpy/_core/include/numpy/` tree (which contains roughly 20 header files, plus the `libdivide` and `random` folders) to NumPy's `cross-build-files`. I decided against doing that, because those files change often between versions and more importantly are not really platform-specific, so listing them would be a maintenance burden and would stretch what our `cross-build-files` field is meant for (as I noted in https://github.com/pyodide/pyodide-recipes/pull/591#issuecomment-4642897159). Instead, we kept the existing step that installs cross-build packages into `HOSTSITEPACKAGES`, so packages that access these files directly from the filesystem, rather than through the isolated build venv, continue to work exactly as before. I expected more packages to have these issues, but only RobotRaconteur and opencv-python did. However, more packages which are based on C/C++ libraries with Python bindings exist in the wild (such as CasADi, which I maintain, though it doesn't need NumPy headers). We need to support such types of builds for packages for better adoption of Pyodide's build machinery.

With that in place, everything builds successfully!


hoodmane:
Sounds good to me. Relying on `PYTHONPATH` being around does indeed seem like it could cause trouble.

pyodide_build/build_env.py
```diff
@@ -169,24 +168,66 @@ def get_hostsitepackages() -> str:
...
+    # TODO: unify libdir for in-tree and out-of-tree builds
+    if in_xbuildenv():
+        libdir = PYODIDE_ROOT / ".." / "site-packages-extras"
+    else:
+        libdir = Path(get_hostsitepackages())
+
+    package_dir = libdir / package_name
+    return libdir, [
+        str(f.relative_to(libdir)) for f in package_dir.rglob("*") if f.is_file()
+    ]
```
hoodmane, line 228: So this is trying to get `cross-build-files`? How does it work in tree? Shouldn't we look at `cross-build-files` in `meta.yaml` in that case?

pyodide_build/tests/test_xbuildenv.py
```diff
@@ -205,33 +205,6 @@ def test_install_force(
         assert (tmp_path / version / ".installed").exists()
         assert manager.current_version == version
 
-    def test_install_cross_build_packages(
```
hoodmane, line 208: It feels like we could use an updated version of this test?

hoodmane (approved):
Thanks @ryanking13! I made some comments but it generally looks good.

pyodide_build/build_env.py
ryanking13: > So this is trying to get cross-build-files? How does it work in tree? Shouldn't we look at cross-build-files in meta.yaml in that case?

As you can see in the `buildpkg.py` change, only cross-build-files are installed under the hostsitepackages on in-tree build (I think there are some room to optimize this logic though).

> Also, site-packages-extras maybe should have been called site-packages-cross-files or something a bit more descriptive...

Totally agree, but the folder `site-packages-extras` is generated when [creating the xbuildenv](https://github.com/pyodide/pyodide/blob/3a6ca52d93153852ada67d3c260594b2f9132f4f/tools/create_xbuildenv.py#L28), so we'll need to fix there too.

agriyakhetarpal:
Hi @ryanking13, I hope you don't mind the recent changes I am making here in the last few commits! I've been hoping to use this to compile SciPy without the f2py generators patch, and I thought the best approach would be to continue your PR rather than creating a new one. That said, it did take me some time to resolve the conflicts, and might I say slightly longer than it would have taken with a fresh PR 😅

pyodide_build/build_env.py
agriyakhetarpal: `site-packages-extras` is well baked into the xbuildenv tarballs at this point. `site-packages-cross-files` would break our backwards compatibility, which we set up after this PR was created (in #167) and have surprisingly managed to maintain it well at this point. I am not opposed to changing things, but since it's separate, I guess it should be out of scope here.

pyodide_build/tests/test_xbuildenv.py
agriyakhetarpal: This comment should no longer be applicable, as the test is back again.

We also now have `test_install_reqs` and `test_replace_unisolated_packages` in `test_pypabuild.py` if you want to take a look.

ryanking13:
@agriyakhetarpal Hey, thanks for working on this. I think this PR can be a stop-gap until numpy and scipy actually starts to publish the pyemscripten wheels to PyPI. Another approach we can think of is #68, which would be a little more robust that it can handle breaking changes in numpy and scipy, but also requires some manual maintenance of those packages.

I am not sure which option is better, but both options are a stop-gap, not a permenant solution, so I am personally okay with either approach.

In the long run, this should not be needed after Numpy and Scipy folks actually start to publish PyEmscripten wheels to PyPI, and people can actually download pyemscripten wheels as a build dependency.

agriyakhetarpal:
> @agriyakhetarpal Hey, thanks for working on this. I think this PR can be a stop-gap until numpy and scipy actually starts to publish the pyemscripten wheels to PyPI. Another approach we can think of is #68, which would be a little more robust that it can handle breaking changes in numpy and scipy, but also requires some manual maintenance of those packages.
> 
> I am not sure which option is better, but both options are a stop-gap, not a permenant solution, so I am personally okay with either approach.
> 
> In the long run, this should not be needed after Numpy and Scipy folks actually start to publish PyEmscripten wheels to PyPI, and people can actually download pyemscripten wheels as a build dependency.

Thanks for your thoughts, @ryanking13! Yes, let me elaborate on these changes a little. I rekindled the fire on this PR for two reasons:
1. Based on [NumPy NEP 57](https://numpy.org/neps/nep-0057-numpy-platform-support.html), which we agreed to be listed as points of contact on, and through the conversations I have had internally at my workplace, it is unlikely that Pyodide will move from Tier 3 up to NumPy Tier 2 in the short or medium term. Tier 3 is currently excluded from publishing to PyPI. I don't wish to imply that we should not ask about it in the NumPy issue tracker at all—we should definitely do so at some point and suggest that they provide PyEmscripten wheels so that our and downstream packagers' lives become easier—it is more that I don't see it happening by the 2027 ABI or so (yet). There are a few good points to note, of course: the fact that PEP 783 has been accepted strengthens our case, as per points 3 and 5 of [NumPy platform support > General principles](https://numpy.org/neps/nep-0057-numpy-platform-support.html#general-principles), and we previously implemented points 2 and 4 when I started working on Pyodide. However, I believe we will fall a bit short on point 1, as none of us are NumPy maintainers 😅
- Making multiple versions of NumPy and SciPy available requires us to set up infrastructure to compile them in GitHub Actions CI (or any other CI provider). This would be done in a repository, where we would need to build at least [NumPy 2.3.2 and above (up to 2.5.0, the latest release)](https://numpy.org/devdocs/release/2.3.2-notes.html), as NumPy 2.3.2 is the earliest version to support Python 3.14. The same would apply to SciPy releases. (We probably won't need to publish all versions for both, since the headers don't change very often, but you get the point :D). Since we have committed to enhancing the security of our supply chain in recent times, I would strongly insist on building and publishing these cross-built wheels through public-facing CI infrastructure with attestations – we could use a temporary repository similar to the one I created at https://github.com/pyodide/pyodide-numpy-2.0-rebuilds or something like https://github.com/pyodide/pyodide-build-environment-nightly. I had looked at #68 previously, and, coupled with these reasons, I believed this PR would require less effort to revive in comparison.

So, while both options are stopgaps indeed, and I would be alright with #68 proceeding as well, I would suggest that we go forward with this one for a start, for the next few months. That said, I would be happy to collaborate with you to set up the alternative build index URL and bring it over the finish line. I believe that would (i) help us let go of the concept of the xbuildenv altogether (or at least slim it down quite a lot as we will still need the Emscripten-compiled Python bits), so there will be follow-up work to do, and (ii) as proper cross-compilation support in the Python packaging ecosystem lands and tools like https://github.com/beeware/xbuild mature, we could look into a larger refactor. Even after #68 were to land, until we let go of the xbuildenv concept and adopt a more conventional approach to building wheels, this PR would still be needed in the short term, as it brings forward the logic to make the console scripts from cross-built packages available; where they come from doesn't change that.

I hope these details help make the case that this PR is a better option (for now) and help guide your judgement! Regarding the robustness of the second approach and the maintenance of NumPy and SciPy to keep them working, I think there are trade-offs on both sides. NumPy has not troubled us recently, thanks to the CI job, but we still need to move miles to compile SciPy with each new version, either way :-)

---

All that noted, https://github.com/pyodide/pyodide-recipes/pull/591 passed the SciPy build with the f2py generators patch removed, and tools like `numpy-config` and `f2py` have been available when building. I confirmed this locally as well. I have triggered a full build on the PR just now to be on the safe side, but otherwise I think I can still mark this as ready for review, though I can't request a review from you directly, as you authored it 😄 Thanks!

agriyakhetarpal:
I'm marking this as a draft for a bit. The approach here works fine for a wide majority of packages when testing in pyodide-recipes, but we still need to be less stringent as it breaks RobotRaconteur and OpenCV: https://github.com/pyodide/pyodide-recipes/pull/591#issuecomment-4642897159

agriyakhetarpal:
Okay, fixed. https://github.com/pyodide/pyodide-recipes/pull/591 is all green now with especially both RobotRaconteur and opencv-python building, and all 325 successful! So I think I should mark this as ready for review 😄

Side note: Claude Sonnet 4.6's Review skill caught a number of interesting small things in the PR, which I fixed, and that I wouldn't have noticed myself.

CHANGELOG.md
```diff
@@ -25,6 +25,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  the cross-build environment. This makes build-time scripts exposed by these
+  packages available on `PATH` during the build process.
+  [#21](https://github.com/pyodide/pyodide-build/pull/21)
+
+- `oldest-supported-numpy` is no longer silently ignored when encountered as a
+  build-time dependency. It will now be installed as any other package. Since
+  `oldest-supported-numpy` is deprecated since NumPy 2.0, packages that still
+  list it should migrate to a direct `numpy` dependency. If you need to ignore
+  it (or any other build requirement), set it via
+  `pyodide config set ignored_build_requirements "patchelf oldest-supported-numpy"`
```
ryanking13: Hallucination from AI agents? I think we never had a `config set` command.

CHANGELOG.md
```diff
@@ -25,6 +25,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   additionally receive a `-v` flag for detailed package resolution output.
   [#363](https://github.com/pyodide/pyodide-build/pull/363)
 
+### Changed
+
+- Unisolated packages are now installed into the isolated build environment via
+  pip, with their cross-build files replaced by WASM-compatible versions from
+  the cross-build environment. This makes build-time scripts exposed by these
+  packages available on `PATH` during the build process.
+  [#21](https://github.com/pyodide/pyodide-build/pull/21)
```
ryanking13: I think we can tidy up this a bit. How isolated packages were installed was not a user-visible behavior so I think we can shorten this a bit, such as:

```
- Fixed build-time scripts of `numpy` not avaialble on `PATH` during the build process
```

integration_tests/recipes/f2py-example/meta.yaml
```diff
@@ -0,0 +1,34 @@
...
+    export F2C_PATH=$(pwd)/f2c/src/f2c
+    cp f2c/f2c.h ${WASM_LIBRARY_DIR}/include/f2c.h
+
+requirements:
+  host:
+    - numpy
+  run:
+    - numpy
+  executable:
+    - gfortran
```
ryanking13: Thanks for writing this integration test. Would it be possible to simplify these a bit? I don't like we need to clone and build f2c just for this, and also requiring host `gfortran` executable.

Also please update https://github.com/pyodide/pyodide-build/blob/main/integration_tests/recipes/README.md mentioning what this recipe is trying to test so that we can reference later.

pyodide_build/tests/test_build_env.py
```diff
@@ -59,14 +59,55 @@ def test_get_pyodide_root_pyodide_root_already_set(
...
+
+    def test_get_unisolated_files(self, dummy_xbuildenv, reset_env_vars, reset_cache):
+        manager = CrossBuildEnvManager(dummy_xbuildenv / common.xbuildenv_dirname())
+        site_packages_extras = manager.pyodide_root / ".." / "site-packages-extras"
+
+        for name in ("numpy", "scipy"):
+            package_dir = build_env.get_unisolated_files(name)
+            assert package_dir == site_packages_extras / name
+            assert package_dir.is_dir()
+            assert any(package_dir.rglob("*"))
```
ryanking13: Not sure if this is a good idea to test this. What packages should exist in the unisolated files can change.

pyodide_build/tests/test_build_env.py
```diff
@@ -59,14 +59,55 @@ def test_get_pyodide_root_pyodide_root_already_set(
...
+
+    def test_get_unisolated_files_no_cross_build_files(
+        self, dummy_xbuildenv, reset_env_vars, reset_cache
+    ):
+        # cffi is an unisolated package but has no cross-build files in the
+        # dummy xbuildenv, so the directory simply doesn't exist
+        package_dir = build_env.get_unisolated_files("cffi")
+        assert not package_dir.exists()
```
ryanking13: This too.

pyodide_build/tests/test_pypabuild.py
```diff
@@ -110,28 +156,102 @@ def test_get_build_env(tmp_path, dummy_xbuildenv):
...
+    purelib.mkdir(parents=True)
+
     monkeypatch.setattr(
-        "pyodide_build.build_env.get_unisolated_packages",
-        lambda: ["numpy"],
+        pypabuild,
+        "_find_executable_and_scripts",
+        lambda venv_path: ("python", "scripts", str(purelib)),
+    )
+    # cffi has no cross-build files, so its directory does not exist
```
ryanking13: This comment is misleading. There is no cross-build file because this test did not set any testfiles, not because cffi has no cross-build files.

pyodide_build/build_env.py
```diff
@@ -220,29 +219,78 @@ def get_hostsitepackages() -> str:
...
+    """
     PYODIDE_ROOT = get_pyodide_root()
 
-    unisolated_file = PYODIDE_ROOT / "unisolated.txt"
-    if unisolated_file.exists():
-        # in xbuild env, read from file
-        unisolated_packages = unisolated_file.read_text().splitlines()
+    unisolated_packages: dict[str, str] = {}
+    if in_xbuildenv():
+        unisolated_packages_file = PYODIDE_ROOT / ".." / "requirements.txt"
```
ryanking13, line 242: Hmm, I don't remember why we had both `unisolated.txt` and `requirements.txt` in xbuildenv... but it looks like `requirements.txt` is more informative.

pyodide_build/pypabuild.py
```diff
@@ -112,27 +114,92 @@ def symlink_unisolated_packages(
...
+    A tuple of (the filtered set of requirements, the set of unisolated requirements)
+    """
+    canonical_unisolated = {
+        canonicalize_name(name): (name, version)
+        for name, version in unisolated_packages.items()
+    }
+
+    new_reqs = reqs.copy()
+    unisolated: set[str] = set()
+    for reqstr in list(reqs):
```
ryanking13: nitpick: set => list conversion looks unnecessary

pyodide_build/pypabuild.py
```diff
@@ -112,27 +114,92 @@ def symlink_unisolated_packages(
...
+        req = Requirement(reqstr)
+        match = canonical_unisolated.get(canonicalize_name(req.name))
+        if match is not None:
```
ryanking13:
```suggestion
        if match is None:
            continue
```

pyodide_build/pypabuild.py
```diff
@@ -112,27 +114,92 @@ def symlink_unisolated_packages(
...
+            name, version = match
+            # TODO: find a better way to handle this case
+            if not req.specifier.contains(version):
+                warnings.warn(
+                    f"Found build dependency {req} but the only supported "
+                    f"cross-build version is {name}=={version}; "
+                    f"using {name}=={version} instead.",
+                    stacklevel=2,
+                )
```
ryanking13: Yeah, we were silently ignoring this version mismatch and showing the warning would be nice.

pyodide_build/build_env.py
```diff
@@ -220,29 +219,78 @@ def get_hostsitepackages() -> str:
...
         recipes = load_all_recipes(recipe_dir)
         for name, config in recipes.items():
             if config.build.cross_build_env:
-                unisolated_packages.append(name)
+                unisolated_packages[name] = config.package.version
 
     return unisolated_packages
 
 
+def get_unisolated_files(package_name: str) -> Path:
```
ryanking13: The function name is a bit misleading. It is actually returning the directory but it reads me like it returns a list of files.

ryanking13 (commented):
Thanks for working on this and sorry for the late review! I left some review but it looks good overall

CHANGELOG.md
agriyakhetarpal: Oof! 😅 Yes, I gave it instructions at the end to write something about `ignored_build_requirements` in `[tool.pyodide.build]`, looks like it misinterpreted me.

pyodide_build/tests/test_pypabuild.py
agriyakhetarpal: Actually, it is partially correct because `cffi` is only included in our xbuildenv's `requirements.txt` (which brings `pycparser` too, as a result) and has no cross-build files. But I get your point 😅 I'll make the change.

pyodide_build/build_env.py
agriyakhetarpal: I've renamed this to `get_cross_build_files_dir`, which seems to make sense.

pyodide_build/build_env.py
agriyakhetarpal: While I was not involved with Pyodide at the time, I think `unisolated.txt` was the simpler version indeed and did not pin the packages. Let me remove that file from the dummy xbuildenv tarball and repack it. Edit: done in a9d4841e518874d4445dec090d8eebe321600e7b.

integration_tests/recipes/f2py-example/meta.yaml
agriyakhetarpal: Yes, makes sense to me. In a1ea4be0ce65354947fa985cf4965a5458dee858, I made it such that `numpy-config` and `f2py` are searched by Meson on PATH, but are not actually needed for the compilation. It is a simple C extension now.

CHANGELOG.md
```diff
@@ -79,6 +79,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   additionally receive a `-v` flag for detailed package resolution output.
   [#363](https://github.com/pyodide/pyodide-build/pull/363)
 
+### Changed
+
+- Fixed build-time scripts of unisolated packages (like `f2py` and
+  `numpy-config` from NumPy) not being available on `PATH` during builds.
+  [#21](https://github.com/pyodide/pyodide-build/pull/21)
+
+- `oldest-supported-numpy` is no longer silently ignored when encountered as a
```
hoodmane: Could we explicitly error on this? It will break later right?

pyodide_build/build_env.py
```diff
@@ -220,29 +219,78 @@ def get_hostsitepackages() -> str:
...
 
-    unisolated_file = PYODIDE_ROOT / "unisolated.txt"
-    if unisolated_file.exists():
-        # in xbuild env, read from file
-        unisolated_packages = unisolated_file.read_text().splitlines()
+    unisolated_packages: dict[str, str] = {}
+    if in_xbuildenv():
+        unisolated_packages_file = PYODIDE_ROOT / ".." / "requirements.txt"
+
+        if unisolated_packages_file.exists():
```
hoodmane: Let's assert that it exists?

pyodide_build/build_env.py
hoodmane: I added requirements.txt with the pins when I switched to xbuildfiles and making the frankenpackages by downloading them from pypi and pasting over the xbuildfiles.

pyodide_build/pypabuild.py
```diff
@@ -112,27 +114,93 @@ def symlink_unisolated_packages(
...
+    unisolated
+        The set of unisolated packages.
+    """
+    if not unisolated:
+        return
+    _, _, purelib = _find_executable_and_scripts(venv_path)
+    sitepackagesdir = Path(purelib)
+    for name in unisolated:
+        package_dir = get_cross_build_files_dir(name)
+        if not package_dir.is_dir():
```
hoodmane, line 186: Why would this happen? Shouldn't we assert that it exists?

CHANGELOG.md
agriyakhetarpal: Yes, we can make it explicitly rejected. I changed it in 38edec2f5374d840a206ab305d64ca74fca388f9.

pyodide_build/build_env.py
agriyakhetarpal: I am not fond of using assert statements outside tests, but yes, we can use a `FileNotFoundError` instead, assuming that's close to what you suggested.

pyodide_build/pypabuild.py
agriyakhetarpal: Not every unisolated package has cross-build files. The package may only need to be pinned to the cross-build version (for its console scripts, for example) without any file overlay. numpy and scipy have cross-build files to be copied, but cffi and pycparser do not. Let me add a comment about this with this explanation.

agriyakhetarpal:
Thanks @hoodmane and @ryanking13 for the reviews! I adjusted the CHANGELOG entries and fixed a few bugs where we were not respecting platform markers when parsing dependencies, which were exposed by raising an explicit error when encountering `oldest-supported-numpy` (Shapely was failing to build). I've updated https://github.com/pyodide/pyodide-recipes/pull/591 and triggered one last test for the full build. We should be good to go ahead with this soon – but please feel free to take another look if you like!

agriyakhetarpal:
Oops, found a problem... but it's SciPy, not us:

SciPy 1.17.1 (when the CI was green in pyodide/pyodide-recipes#591) had `dependency('numpy', required: false)`

SciPy 1.18.0 has `dependency('numpy')` and makes it explicitly required.

agriyakhetarpal:
So I think we can either pass another cross file, in flight (Meson supports multiple cross files), which would look like:

```ini
[binaries]
numpy-config = '/tmp/build-env-abcdef/bin/numpy-config'
f2py = '...'
```

and this won't be exposed to the user (but I don't think it's an elegant solution).

Or (I chose this) we can use `pkg-config` and pass `PKG_CONFIG_LIBDIR`, since NumPy ships a `.pc` file. My reason for doing so is https://github.com/scipy/scipy/pull/24119#discussion_r2614925746, but please let me know if we should ask Ralf to revisit this.

agriyakhetarpal:
Updated fiona as it was using `oldest-supported-numpy`; hopefully there are no others. Building all recipes locally is a pain, so I am relying on CI to do this.

Edit: https://github.com/pyodide/pyodide-recipes/pull/591 is all green now. Fiona was the only package using `oldest-supported-numpy`.

pyodide_build/pypabuild.py
ryanking13: Yeah, I think it is good to be extra safe here as the format of xbuildenv files can change in the future, and we can make mistakes.

pyodide_build/pypabuild.py
```diff
@@ -112,29 +114,108 @@ def symlink_unisolated_packages(
...
+    canonical_unisolated = {
+        canonicalize_name(name): (name, version)
+        for name, version in unisolated_packages.items()
+    }
+
+    new_reqs = reqs.copy()
+    unisolated: set[str] = set()
+    for reqstr in reqs:
+        req = Requirement(reqstr)
+        if req.marker and not req.marker.evaluate():
```
ryanking13, line 147: could please add a comment of what `not req.marker.evaulate()` mean?

pyodide_build/pypabuild.py
```diff
@@ -112,29 +114,108 @@ def symlink_unisolated_packages(
...
+            continue
+        if canonicalize_name(req.name) == "oldest-supported-numpy":
```
ryanking13: sounds okay to me, but could you please add a comment why we are not supporting it?

pyodide_build/pypabuild.py
```diff
@@ -147,16 +228,32 @@ def install_reqs(
...
-            )
-        )
+        env.install(reqs)
+
+    _install_cross_build_files(env.path, unisolated)
+
+
+# So far among all packages in pyodide-recipes, only NumPy ships
+# a .pc file, but I don't want to hardcode that here as such
+def _get_unisolated_pkgconfig_dirs(venv_path: str) -> list[str]:
```
ryanking13, line 246: Makes sense to me. I am curious how numpy adds these files to PKG_CONFIG_LIBDIR during the build

ryanking13 (commented):
It looks like I cannot approve this as this was originally my PR, but you can consider this comment as my approval.

Thanks for reviving this PR!

pyodide_build/pypabuild.py
agriyakhetarpal: The only reason for not supporting it is because @hoodmane asked 😄 https://github.com/pyodide/pyodide-build/pull/21#discussion_r3485877731. It is not strictly needed for this PR, though. It won't break anything with the functionality change here; it's more of a deadweight cleanup. I can split it into its own PR.

pyodide_build/pypabuild.py
agriyakhetarpal: Actually, NumPy's only task is to ship the `.pc` file. That is recognised by Meson during the dependency (`dependency('numpy')`) search as one of the files to look on the system for. We then point the `PKG_CONFIG_LIBDIR` environment variable to it ourselves. 

Another way for Meson to find NumPy is to list `numpy-config` in the `[binaries]` section of the Meson cross file. Right now:

```
numpy-config binary missing from cross or native file, or env var undefined.
Default target is not allowed for cross use
numpy-config found: NO
```

Meson does find numpy-config, but it refuses to use it because it's a host-machine tool. Meson assumes it may contain incorrect include/library paths for WASM, and we need to alter the cross file and add `numpy-config` to make it not do so.

However, I felt the pkg-config approach was cleaner, and that's what was recommended, so I went ahead with it.

pyodide_build/pypabuild.py
agriyakhetarpal: Yes, I'll add one. Noting it here where I encountered it: our current Shapely recipe here uses `oldest-supported-numpy` as a build-time dependency only for Python versions older than 3.9. So without `evaluate()`ing the PEP 508 markers, we would look at `oldest-supported-numpy ; python_version < '3.9'`, find only `oldest-supported-numpy` and trip up, raising an error even though it shouldn't concern us..

agriyakhetarpal:
Thanks for reviewing, both of you!

PR #27, opened 2024-09-11
Include version and checksum validation when updating packages

### LINKED ISSUE #24: `pyodide skeleton pypi --update` does not completely validate checksums when updating
## Description

The `pyodide skeleton pypi` commands (whether `--update` or `--update-patched`) won't update the hash if they infer the package version is already updated.

## Reproducer

This is a special case where one manually updates the `version:` number for `mypackage` in its recipe and leaves the hash unchanged in the `sha256:` field. In this case, running `pyodide skeleton pypi --update mypackage` will not update the hash, assuming that the user has updated it and validated it themselves. Here's a small example:

<details>
<summary>Click to expand code</summary>

#### `numpy/meta.yaml` (before)

```yaml
package:
  name: numpy
  version: 2.0.1
  tag:
    - min-scipy-stack
  top-level:
    - numpy
source:
  sha256: randomhash12345
```

#### `numpy/meta.yaml` (after a manual bump to the version – 2.1.1 is the latest one available on PyPI at the time of writing)

```yaml
package:
  name: numpy
  version: 2.1.1
  tag:
    - min-scipy-stack
  top-level:
    - numpy
source:
  sha256: randomhash12345
```

Now, if one runs `pyodide skeleton pypi --update numpy`, they get

#### `numpy/meta.yaml` has no change:

```yaml
package:
  name: numpy
  version: 2.1.1
  tag:
    - min-scipy-stack
  top-level:
    - numpy
source:
  sha256: randomhash12345
```

``` bash
❯ pyodide skeleton pypi --update numpy
numpy already up to date. Local: 2.1.1 PyPI: 2.1.1
```

</details>

## Expected behaviour

I think the better pathway would be to assume that a user or recipe maintainer can make mistakes when updating package versions manually, but a computer would not – so, I feel that the command should attempt to update the checksum anyway, even if the version matches the one available on PyPI. This should be the same for `--update` and `--update-patched`.

## Additional information

N/A

agriyakhetarpal:
This PR closes #24. Here, we now check against two conditions when updating the package
- whether the version is up to date plus valid, and
- whether the SHA-256 checksum matches the one present in the recipe file.

Therefore, updating the package has three scenarios, and here is a summary of what happens:
- local version updated manually to the newest available PyPI version ➡️ checksums will be updated, too
- both version and checksum are out of date ➡️ both of them will be updated (as it was before these changes)
- the checksum is correct and the version is out of date ➡️ this is enough information for us to proceed with updating the version

Additionally, there's now a case where one manually updates the version (in error, for example), to a version that is not released or available on PyPI (yet), which means that the metadata for the version won't exist either. We raise an exception early here, asking the user so that they can check the version while updating. This is a rare situation, so it's more about raising a helpful error that aids the user in debugging the problem. I have skipped adding a test for such a case, but I can add one if needed. Please let me know your thoughts!

### DIFF
--- a/pyodide_build/mkpkg.py
+++ b/pyodide_build/mkpkg.py
@@ -256,6 +256,15 @@ def update_package(
     sha256 = dist_metadata["digests"]["sha256"]
     sha256_local = yaml_content["source"].get("sha256")
 
+    # fail if local version is newer than PyPI version
+    # since updating isn't possible in that case
+    if pypi_ver <= local_ver:
+        raise MkpkgFailedException(
+            f"Local version {local_ver} is newer than PyPI version {pypi_ver}, "
+            f"cannot update {package}. Please verify in case the version was "
+            "updated manually and is correct."
+        )
+
     # conditions to check if the package is up to date
     is_sha256_up_to_date = sha256 == sha256_local
     is_version_up_to_date = pypi_ver <= local_ver
@@ -266,7 +275,7 @@ def update_package(
     if already_up_to_date:
         logger.success(
             f"{package} already up to date."
-            f" Local: {local_ver} PyPI: {pypi_ver}"
+            f" Local: {local_ver} and PyPI: {pypi_ver}"
             f" and checksum received: {sha256} matches local: {sha256_local} ✅"
         )
         return
--- a/pyodide_build/mkpkg.py
+++ b/pyodide_build/mkpkg.py
@@ -245,18 +245,37 @@ def update_package(
         old_fmt = "sdist"
 
     pypi_metadata = _get_metadata(package, version)
+
+    # Grab versions from metadata
     pypi_ver = Version(pypi_metadata["info"]["version"])
     local_ver = Version(yaml_content["package"]["version"])
-    already_up_to_date = pypi_ver <= local_ver and (
+
+    # and grab checksums from metadata
+    source_fmt = source_fmt or old_fmt
+    dist_metadata = _find_dist(pypi_metadata, [source_fmt])
+    sha256 = dist_metadata["digests"]["sha256"]
+    sha256_local = yaml_content["source"].get("sha256")
+
+    # conditions to check if the package is up to date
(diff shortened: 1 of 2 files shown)


agriyakhetarpal:
Some tests always fail for me locally due to differences in errors from assert statements, but this is probably because my default shell is not `bash`. I think these changes should be fine here, but I'll trigger the integration tests so that we are sure.

agriyakhetarpal:
I guess coverage isn't configured yet, it's the same for other PRs, too.

pyodide_build/mkpkg.py
```diff
@@ -245,18 +245,46 @@ def update_package(
...
+    if pypi_ver <= local_ver:
+        raise MkpkgFailedException(
+            f"Local version {local_ver} is newer than PyPI version {pypi_ver}, "
+            f"cannot update {package}. Please verify in case the version was "
+            "updated manually and is correct."
+        )
+
+    # conditions to check if the package is up to date
+    is_sha256_up_to_date = sha256 == sha256_local
+    is_version_up_to_date = pypi_ver <= local_ver
```
ryanking13, line 270: When `pypi_ver <= local_ver` is true, it fails with exception, but `is_version_up_to_date` checks the same condition after the exception.

ryanking13 (commented):
Thanks @agriyakhetarpal! We haven't been maintaining `skeleton` command that much, so it is nice you are working on it.

pyodide_build/mkpkg.py
agriyakhetarpal: Oops! Thanks for noticing. 143b729360f48cc912b634ffcc1c72c7ffd2463a fixes this – the version comparison should be made in equality (`pypi_ver == local_ver`), and the impossible update scenario should have a strict greater-than (`pypi_ver < local_ver`).

ryanking13 (approved):
Thanks @agriyakhetarpal!

agriyakhetarpal:
Thanks for the review!

PR #28, opened 2024-09-11
Add a JSON output for `pyodide xbuildenv search`, better tabular output

### LINKED ISSUE #26: `pyodide xbuildenv search --json` option
## Description

The `pyodide xbuildenv search` command currently prints out a formatted table of values:

```bash
❯ pyodide xbuildenv search
```

currently returns

```
Starting new HTTPS connection (1): raw.githubusercontent.com:443
https://raw.githubusercontent.com:443 "GET /pyodide/pyodide/main/pyodide-cross-build-environments.json HTTP/11" 200 917
Version         Python          Emscripten      pyodide-build                   Compatible
----------      ----------      ----------      -------------------------       ----------
0.27.0a2        3.12.1          3.1.58          0.26.0 -                        Yes       
0.26.2          3.12.1          3.1.58          0.26.0 -                        Yes       
0.26.1          3.12.1          3.1.58          0.26.0 -                        Yes       
0.26.0          3.12.1          3.1.58          0.26.0 -                        Yes    
```

(Side note: is the `pyodide-build` version wrong? It's returning ` 0.26.0 -` for all the versions, and not the latest `pyodide-build` version at the time of writing – which is version 0.28.0?)

It would be great to get the same output in JSON – for example, something like

```json
{
  "environments": [
    {
      "version": "0.27.0a2",
      "python": "3.12.1",
      "emscripten": "3.1.58",
      "pyodide_build": "0.26.0",
      "compatible": true
    },
    {
      "version": "0.26.2",
      "python": "3.12.1",
      "emscripten": "3.1.58",
      "pyodide_build": "0.26.0",
      "compatible": true
    },
    {
      "version": "0.26.1",
      "python": "3.12.1",
      "emscripten": "3.1.58",
      "pyodide_build": "0.26.0",
      "compatible": true
    },
    {
      "version": "0.26.0",
      "python": "3.12.1",
      "emscripten": "3.1.58",
      "pyodide_build": "0.26.0",
      "compatible": true
    }
  ]
}
```

## Motivation

A JSON-based output would be machine-readable and serialisable into a Python script at runtime. This might be useful for `cibuildwheel`, where currently it is assumed that the Pyodide version and the `pyodide-build` version go hand-in-hand (it has to be updated since the unvendoring, which I'm working on).

agriyakhetarpal:
## Description

This PR closes #26. It adds a `pyodide xbuildenv search --json` option to print a JSON-based output, along with associated tests. The advantage is that it can be saved to a file, piped to `jq` or shell functions (or any equivalent tools), or simply imported into a Pythonic interface.

Additionally, I added a small context manager that does not do anything but stop printing the following lines:
```
Starting new HTTPS connection (1): raw.githubusercontent.com:443
https://raw.githubusercontent.com:443 "GET /pyodide/pyodide/main/pyodide-cross-build-environments.json HTTP/11" 200 917
```

in the output, because it conflicts with receiving valid JSON. Please let me know if this would be undesirable. If yes, I'll try to work around it so that it gets printed for the non-JSON output (table).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - The `pyodide xbuildenv search` command now accepts a `--json` flag to output the
-  search results in JSON format that is machine-readable.
+  search results in JSON format that is machine-readable. The design for the regular
+  tabular output has been improved.
   [#28](https://github.com/pyodide/pyodide-build/pull/28)
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- The `pyodide xbuildenv search` command now accepts a `--json` flag to output the
+  search results in JSON format that is machine-readable.
+  [#28](https://github.com/pyodide/pyodide-build/pull/28)
+
 - `pyo3_config_file` is no longer available in `pyodide config` command.
   Pyodide now sets `PYO3_CROSS_PYTHON_VERSION`, `PYO3_CROSS_LIB_DIR` to specify the cross compilation environment
   for PyO3.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-### Changed
+### Added
 
 - The `pyodide xbuildenv search` command now accepts a `--json` flag to output the
   search results in JSON format that is machine-readable.
   [#28](https://github.com/pyodide/pyodide-build/pull/28)
+
+### Changed
 
 - `pyo3_config_file` is no longer available in `pyodide config` command.
   Pyodide now sets `PYO3_CROSS_PYTHON_VERSION`, `PYO3_CROSS_LIB_DIR` to specify the cross compilation environment
@@ -30,7 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.27.2] - 2024/07/11
 
-## Changed
+### Changed
 
 - `pyodide py-compile` command now accepts `excludes` flag.
   [#9](https://github.com/pyodide/pyodide-build/pull/9)
@@ -40,7 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.27.1] - 2024/06/28
 
-## Changed
+### Changed
 
 - ported f2c_fixes patch from https://github.com/pyodide/pyodide/pull/4822
 
--- a/pyodide_build/cli/xbuildenv.py
+++ b/pyodide_build/cli/xbuildenv.py
@@ -4,6 +4,7 @@ import typer
 
 from ..build_env import local_versions
 from ..common import xbuildenv_dirname
+from ..views import MetadataView
 from ..xbuildenv import CrossBuildEnvManager
 from ..xbuildenv_releases import (
     cross_build_env_metadata_url,
@@ -180,103 +181,25 @@ def _search(
         )
         raise typer.Exit(1)
 
-    def _generate_json_output(releases, local) -> str:
-        """A helper function to help generate JSON output"""
-        import json
-
-        output = {
-            "environments": [
-                {
-                    "version": release.version,
-                    "python": release.python_version,
-                    "emscripten": release.emscripten_version,
-                    "pyodide_build": {
-                        "min": release.min_pyodide_build_version,
-                        "max": release.max_pyodide_build_version,
-                    },
-                    "compatible": release.is_compatible(
-                        python_version=local["python"],
-                        pyodide_build_version=local["pyodide-build"],
-                    ),
-                }
-                for release in releases
-            ]
-        }
-        print(json.dumps(output, indent=2))
-
-    def _print_table_output(releases, local) -> None:
-        """A helper function to print a tabular output"""
-
-        columns = [
-            ("Version", 10),
-            ("Python", 10),
-            ("Emscripten", 10),
-            ("pyodide-build", 25),
-            ("Compatible", 10),
-        ]
-
-        # Unicode box-drawing characters
-        top_left = "┌"
-        top_right = "┐"
-        bottom_left = "└"
-        bottom_right = "┘"
-        horizontal = "─"
-        vertical = "│"
-        t_down = "┬"
-        t_up = "┴"
-        t_right = "├"
-        t_left = "┤"
-        cross = "┼"
-
-        # Table elements
-        top_border = (
-            top_left
-            + t_down.join(horizontal * (width + 2) for _, width in columns)
-            + top_right
+    # Generate views for the metadata objects (currently tabular or JSON)
+    views = [
+        MetadataView(
+            version=release.version,
+            python=release.python_version,
+            emscripten=release.emscripten_version,
+            pyodide_build={
+                "min": release.min_pyodide_build_version,
+                "max": release.max_pyodide_build_version,
+            },
+            compatible=release.is_compatible(
+                python_version=local["python"],
+                pyodide_build_version=local["pyodide-build"],
+            ),
         )
-        header = (
-            vertical
-            + vertical.join(f" {name:<{width}} " for name, width in columns)
-            + vertical
-        )
-        separator = (
-            t_right
-            + cross.join(horizontal * (width + 2) for _, width in columns)
-            + t_left
-        )
-        bottom_border = (
-            bottom_left
-            + t_up.join(horizontal * (width + 2) for _, width in columns)
-            + bottom_right
-        )
-
-        ### Printing
-        print(top_border)
-        print(header)
-        print(separator)
-        for release in releases:
-            compatible = (
-                "Yes"
-                if release.is_compatible(
-                    python_version=local["python"],
-                    pyodide_build_version=local["pyodide-build"],
-                )
-                else "No"
-            )
-            pyodide_build_range = f"{release.min_pyodide_build_version or ''} - {release.max_pyodide_build_version or ''}"
-
-            row = [
-                f"{release.version:<{columns[0][1]}}",
-                f"{release.python_version:<{columns[1][1]}}",
-                f"{release.emscripten_version:<{columns[2][1]}}",
-                f"{pyodide_build_range:<{columns[3][1]}}",
-                f"{compatible:<{columns[4][1]}}",
-            ]
-
-            print(vertical + vertical.join(f" {cell} " for cell in row) + vertical)
-        print(bottom_border)
+        for release in releases
+    ]
 
     if json_output:
-        _generate_json_output(releases, local)
+        MetadataView.to_json(views)
     else:
-        _print_table_output(releases, local)
+        MetadataView.to_table(views)
--- a/pyodide_build/cli/xbuildenv.py
+++ b/pyodide_build/cli/xbuildenv.py
@@ -200,6 +200,6 @@ def _search(
     ]
 
     if json_output:
-        MetadataView.to_json(views)
+        print(MetadataView.to_json(views))
     else:
-        MetadataView.to_table(views)
+        print(MetadataView.to_table(views))
--- a/pyodide_build/cli/xbuildenv.py
+++ b/pyodide_build/cli/xbuildenv.py
@@ -156,7 +156,7 @@ def _search(
         "--json",
         help="output results in JSON format",
     ),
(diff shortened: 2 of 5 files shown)


agriyakhetarpal:
Also, I got slightly nerdswiped by the tabular outputs we have, and implemented a prettified version of it (from https://ozh.github.io/ascii-tables/) in ce1904a5e0003d96e3c5149829e9c4ab2f164047, see below:

#### `pyodide xbuildenv search` on the `main` branch

```
Version         Python          Emscripten      pyodide-build                   Compatible
----------      ----------      ----------      -------------------------       ----------
0.27.0a2        3.12.1          3.1.58          0.26.0 -                        Yes       
0.26.2          3.12.1          3.1.58          0.26.0 -                        Yes       
0.26.1          3.12.1          3.1.58          0.26.0 -                        Yes       
0.26.0          3.12.1          3.1.58          0.26.0 -                        Yes       
```

#### `pyodide xbuildenv search` with Unicode symbols

```
┌────────────┬────────────┬────────────┬───────────────────────────┬────────────┐
│ Version    │ Python     │ Emscripten │ pyodide-build             │ Compatible │
├────────────┼────────────┼────────────┼───────────────────────────┼────────────┤
│ 0.27.0a2   │ 3.12.1     │ 3.1.58     │ 0.26.0 -                  │ Yes        │
│ 0.26.2     │ 3.12.1     │ 3.1.58     │ 0.26.0 -                  │ Yes        │
│ 0.26.1     │ 3.12.1     │ 3.1.58     │ 0.26.0 -                  │ Yes        │
│ 0.26.0     │ 3.12.1     │ 3.1.58     │ 0.26.0 -                  │ Yes        │
└────────────┴────────────┴────────────┴───────────────────────────┴────────────┘
```

which is quite nicer!

agriyakhetarpal:
The only thing left for me to do here is check with `cibuildwheel` and see if this works there or if we need something extra. I'll do that later in the day today and mark this ready for review.

agriyakhetarpal:
I've tried this with `cibuildwheel`, and it should work fine. The build procedure in `cibuildwheel` is a bit complicated for me to understand, but the idea is that I'm successfully able to retrieve a list of compatible Pyodide xbuildenv versions within Python as a list of strings: `["0.27.0a2", "0.26.2", "0.26.1", "0.26.0"]` from the stdout stream from running this command.

This should help set up a base for developing wheels against arbitrary Pyodide versions in `cibuildwheel`. After we bump the constraints to use a newer version of `pyodide-build` in `cibuildwheel/resources/constraints-pyodide312.txt`, we'll be able to retrieve the above output from the JSON, and pass a non-alpha Pyodide xbuildenv version to the `install_xbuildenv()` method in `cibuildwheel/pyodide.py`. @ryanking13, I do have a question about this approach, though: how should we decide on what Pyodide versions to support? i.e., if Pyodide v0.X.Y would be available through `pyodide-build` v0.M.N, what would be the suitable values of X, Y, M, and N? In real-world scenarios, the question would be whether `pyodide-build` v0.35.0 in the future can support, say, a much older Pyodide 0.26.0? In case we implement some breaking changes in `pyodide-build`, we would have to bump the minimum version of Pyodide that we are compatible with. Do we have a policy on that, thus far? If not, should we implement one?

In any case, this is ready for review – if we need a better output from this command, we can always implement improvements to it in subsequent PRs and versions.

agriyakhetarpal:
Also, the context manager fixes our problem that was present in https://github.com/pyodide/pyodide/pull/4814.

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- The `pyodide xbuildenv search` command now accepts a `--json` flag to output the
+  search results in JSON format that is machine-readable.
+  [#28](https://github.com/pyodide/pyodide-build/pull/28)
```
ryanking13, line 14: Let's make an [`Added` Section](https://keepachangelog.com/en/1.0.0/#how) and put it in that section.

ryanking13:
> I do have a question about this approach, though: how should we decide on what Pyodide versions to support? i.e., if Pyodide v0.X.Y would be available through pyodide-build v0.M.N, what would be the suitable values of X, Y, M, and N? In real-world scenarios, the question would be whether pyodide-build v0.35.0 in the future can support, say, a much older Pyodide 0.26.0? In case we implement some breaking changes in pyodide-build, we would have to bump the minimum version of Pyodide that we are compatible with. Do we have a policy on that, thus far? If not, should we implement one?

My goal is to prevent breaking changes as much as possible, but yes, some changes are inevitable.

If any breaking change happens, I'll basically update the `min_pyodide_build_version`, and `max_pyodide_build_version` in the [metadata file](https://github.com/pyodide/pyodide/blob/main/pyodide-cross-build-environments.json), to specify the compatibility of each Pyodide version and pyodide-build.

The problem at this point is that when people say they want to build a package that targets an older version of Pyodide, they don't know which version of pyodide-build to install.  What people can do now is to install pyodide-build first, check for compatible versions with `pyodide xbuildenv search`, and then install pyodide-build for those versions again, which is not a very good user experience. I haven't come up with a clean solution to this yet, so if you have any ideas, please share it with us.

pyodide_build/cli/xbuildenv.py
```diff
@@ -175,40 +180,103 @@ def _search(
...
-        )
-        pyodide_build_range = f"{release.min_pyodide_build_version or ''} - {release.max_pyodide_build_version or ''}"
-
-        row = [
-            f"{release.version:{columns[0][1]}}",
-            f"{release.python_version:{columns[1][1]}}",
-            f"{release.emscripten_version:{columns[2][1]}}",
-            f"{pyodide_build_range:{columns[3][1]}}",
-            f"{compatible:{columns[4][1]}}",
+    def _generate_json_output(releases, local) -> str:
```
ryanking13, line 183: How about extracting out `_generate_json_output`, `_print_table_output` into a class and put those function as a method. So we can make the function smaller and make it more testable. Such as:

```python
@dataclass
class MetadataView:
  version: ...
  python: ...
  emscripten: ...
  pyodide_build: ...
  compatible: ...

  def to_json():
        ...

  def to_table():
        ...
```

ryanking13 (commented):
Thanks! I left some comment.

agriyakhetarpal:
> What people can do now is to install pyodide-build first, check for compatible versions with `pyodide xbuildenv search`, and then install pyodide-build for those versions again, which is not a very good user experience. I haven't come up with a clean solution to this yet, so if you have any ideas, please share it with us.

I have a few thoughts in mind:
- One way could be to backport the xbuildenv search feature to older `pyodide-build` versions. This is doable, but it comes at the cost of doing it and backporting it, which is tedious.
- We can't really support more than one Python version in Pyodide,  because doing so would be a burden on maintenance.
- We could add the generated table of supported `Pyodide` versions against a `pyodide-build` version to the README here and continually update it after every release and/or breaking change. This would benefit users because they would know where to look for what version to install directly without having to download and execute code.
- Another aspect is that a lot of people are going to use `cibuildwheel` to build Pyodide binary wheels in conjunction with already established, `cibuildwheel`-enabled, CD jobs, and once we have a reasonable wheel hosting solution, upload the wheel artifacts there. So, it would make sense for `cibuildwheel` to do this internally and suggest a few Pyodide versions to the users against the constrained `pyodide-build` version that would be present there (which

I think it all boils down to the Pyodide version support we can offer in a realistic manner. For example, we should make it explicit that someone should not use, say, Pyodide version 0.21 in 2024 (just an example). The Pyodide version is also influenced by the versions of the packages present in it, because they would have to be compatible with the CPython version in Pyodide. This symbiotic relationship between the Pyodide, CPython, and the Emscripten versions is what makes Pyodide work as a full-fledged distribution, but it doesn't offer guarantees on release and support cycles. With the Scientific Python ecosystem having adopted [SPEC 0000](https://scientific-python.org/specs/spec-0000/) for bumping minimum versions of dependencies and minimum bounds of CPython, if we have our own consensus on a minimum supported Pyodide version in the release cycle (please let me know if we have one already!), we can better gauge a solution for `pyodide-build`'s support for a minimum Pyodide version.

Either way, a Markdown table in the README could be added (from the third point).

CHANGELOG.md
agriyakhetarpal: Makes sense, addressed in f8f639697f7d873c57dd15f32304d2ace974d0dd

agriyakhetarpal: Makes sense, addressed in f8f639697f7d873c57dd15f32304d2ace974d0dd.

agriyakhetarpal:
The integration tests seem to fail to run me on my fork, but that is not the case here. I triggered them here so that there are no regressions later on.

ryanking13:
> We could add the generated table of supported Pyodide versions against a pyodide-build version to the README here and continually update it after every release and/or breaking change. This would benefit users because they would know where to look for what version to install directly without having to download and execute code.

Yes, I think this is a most doable solution that does not increase the maintenance burden. We can probably also provide a file that specifies the maximum compatible pyodide-build version for each version of Pyodide (actually, our metadata file can already be used for that feature).

> if we have our own consensus on a minimum supported Pyodide version in the release cycle

Actually, we don't have any support for older Pyodide versions. If something breaks in older Pyodide, we just ask them to "update the Pyodide" (or pyodide-build). But yes, as now pyodide-build is unvendored and people start to build packages with cibuildwheel, we need to consider it...

pyodide_build/views.py
```diff
@@ -0,0 +1,92 @@
...
+                f"{view.python:<{columns[1][1]}}",
+                f"{view.emscripten:<{columns[2][1]}}",
+                f"{pyodide_build_range:<{columns[3][1]}}",
+                f"{'Yes' if view.compatible else 'No':<{columns[4][1]}}",
+            ]
+            table.append(
+                vertical + vertical.join(f" {cell} " for cell in row) + vertical
+            )
+        table.append(bottom_border)
+        print("\n".join(table))
```
ryanking13, line 73: Let's return the string not printing the value, which is easier to test.

pyodide_build/views.py
```diff
@@ -0,0 +1,92 @@
...
+                        "emscripten": view.emscripten,
+                        "pyodide_build": view.pyodide_build,
+                        "compatible": view.compatible,
+                    }
+                    for view in views
+                ]
+            },
+            indent=2,
+        )
+        print(result)
```
ryanking13, line 92: Here too.

pyodide_build/tests/test_cli_xbuildenv.py
```diff
@@ -25,6 +26,18 @@ def mock_pyodide_lock() -> PyodideLockSpec:
     )
 
 
+@pytest.fixture()
+def is_valid_json():
+    def _is_valid_json(json_str):
```
ryanking13, line 31: Question: Why is it defined as a fixture? Doesn't calling `is_valid_json` work?

ryanking13 (approved):
Thanks! I left some minor comments.

pyodide_build/views.py
agriyakhetarpal: Thanks, addressed in 5370c25d9da737d97d1152fda35c542b4c408ac3.

pyodide_build/views.py
agriyakhetarpal: Thanks, addressed in 5370c25d9da737d97d1152fda35c542b4c408ac3.

pyodide_build/tests/test_cli_xbuildenv.py
agriyakhetarpal: Oh, I was using it in the test invocation, which meant that it needed to be a fixture for `pytest` to recognise it. But we don't need that, necessarily. In a9f9a61ee3936f05f7018f6491f934de09950a9d, I've removed the fixture and called it directly, as you suggest.

agriyakhetarpal:
> > We could add the generated table of supported Pyodide versions against a pyodide-build version to the README here and continually update it after every release and/or breaking change. This would benefit users because they would know where to look for what version to install directly without having to download and execute code.
> 
> Yes, I think this is a most doable solution that does not increase the maintenance burden. We can probably also provide a file that specifies the maximum compatible pyodide-build version for each version of Pyodide (actually, our metadata file can already be used for that feature).
> 
> > if we have our own consensus on a minimum supported Pyodide version in the release cycle
> 
> Actually, we don't have any support for older Pyodide versions. If something breaks in older Pyodide, we just ask them to "update the Pyodide" (or pyodide-build). But yes, as now pyodide-build is unvendored and people start to build packages with cibuildwheel, we need to consider it...

Thanks for your thoughts on this. Yes, linking to the metadata file makes sense, too (it's just that a formatted table is more readable with the same data). I guess we should open a new issue about this, because it is also tied in with, say, including multiple versions of packages in the same Pyodide version (useful for interactive documentation across versions). I guess I need to write a detailed RFC issue about that prospect...

agriyakhetarpal:
I'll put up a reminder about thinking more about this, I'll hopefully write something after PyCon this week. In other words, I guess this is ready to merge, now. Thank you for the review!

PR #29, opened 2024-09-16
chore: convert all relative imports to absolute imports

agriyakhetarpal:
## Description

This is a small PR that does some cleanup I wanted to do for a while:

- Converted all relative imports to absolute imports
- Added a rule against this: https://docs.astral.sh/ruff/rules/relative-imports/

Relative imports are not bad, it's just that absolute imports are nicer, at least to me :)

### DIFF
--- a/pyodide_build/_py_compile.py
+++ b/pyodide_build/_py_compile.py
@@ -11,8 +11,8 @@ from typing import Any
 from packaging.tags import Tag
 from packaging.utils import parse_wheel_filename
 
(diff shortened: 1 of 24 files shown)


ryanking13 (approved):
Thanks! I often use relative imports just to not repeat the package name everywhere, but I don't have much preference, so I am okay with this change.

agriyakhetarpal:
That's a valid reason to use them, too!

PR #30, opened 2024-09-17
Do not require downloading xbuildenv when updating package recipes

### LINKED ISSUE #25: `pyodide skeleton pypi --update` (or `--update-patched`) shouldn’t need to install xbuildenv if it fails
## Description

I'm not sure if this is expected behaviour or not, but the task of updating a package version using these commands downloads and sets up an xbuildenv before trying to update. This is not caught in a Docker container or in the Pyodide repository folder, because the xbuildenv exists there anyway, and I understand that there is currently no use for this command outside of those locations. Thus, while it's probably not very relevant – this could be a potential improvement towards demarcating CLI commands into use cases that are distinct from each other. 

However, it could prove useful once we complete the unvendoring of the package recipes into a separate `pyodide-recipes` organisation/repository (see pyodide/pyodide#4918 and pyodide/pyodide#4987).

## Expected behaviour

Since the xbuildenv is not required to update packages, we could refactor our code to not install it at that moment.

agriyakhetarpal:
## Description

Closes #25. This PR improves the xbuildenv search logic to not require downloads of the xbuildenv if a package update fails (for any reason). This means that the Pyodide root is searched in the current directory, and if it doesn't exist, we simply assume that the current working directory is the Pyodide root (and the `packages/` directory one level down hosts the recipes).

I confirmed that this does not currently cause issues with the existing setup in the Pyodide repository.

### DIFF
--- a/pyodide_build/cli/skeleton.py
+++ b/pyodide_build/cli/skeleton.py
@@ -76,7 +76,7 @@ def new_recipe_pypi(
             logger.error(f"{name} update failed: {e}")
             sys.exit(1)
         except mkpkg.MkpkgSkipped as e:
-            logger.warn(f"{name} update skipped: {e}")
+            logger.warning(f"{name} update skipped: {e}")
         except Exception:
             print(name)
             raise
--- a/pyodide_build/cli/skeleton.py
+++ b/pyodide_build/cli/skeleton.py
@@ -51,16 +51,23 @@ def new_recipe_pypi(
     Create a new package recipe from PyPI or update an existing recipe.
     """
 
+    # Determine the recipe directory. If it is specified by the user, we use that;
+    # otherwise, we assume that the recipe directory is the ``packages`` directory
+    # in the root of the Pyodide tree, without the need to initialize the
+    # cross-build environment.
+    #
+    # It is unlikely that a user will run this command outside of the Pyodide
+    # tree, so we do not need to initialize the environment at this stage.
+
+    cwd = Path.cwd()
+    root = build_env.search_pyodide_root(curdir=cwd)
+
+    if not root:
+        root = cwd
+
     if recipe_dir:
         recipe_dir_ = Path(recipe_dir)
     else:
-        cwd = Path.cwd()
-
-        if build_env.in_xbuildenv():
-            root = cwd
-        else:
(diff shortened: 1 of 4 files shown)


pyodide_build/cli/skeleton.py
```diff
@@ -48,19 +48,26 @@ def new_recipe_pypi(
...
+    # cross-build environment.
+    #
+    # It is unlikely that a user will run this command outside of the Pyodide
+    # tree, so we do not need to initialize the environment at this stage.
+
+    cwd = Path.cwd()
+    root = build_env.search_pyodide_root(curdir=cwd)
+
+    if not root:
+        root = cwd
```
ryanking13, line 66: We don't need to calculate these when recipe_dir is specified, so how about putting it in the else block?

pyodide_build/xbuildenv.py
```diff
@@ -250,7 +250,7 @@ def _download(self, url: str, path: Path) -> None:
             Path to extract the cross-build environment to.
             If the path already exists, raise an error.
         """
-        logger.info("Downloading Pyodide cross-build environment from %s", url)
+        logger.info(f"Downloading Pyodide cross-build environment from {url}")
```
ryanking13, line 253: I recently learned that it is not recommended not to use f-strings in log calls (https://github.com/pyodide/micropip/pull/131).

ryanking13 (approved):
Thanks! I left some minor comments, otherwise looks good to me.

pyodide_build/xbuildenv.py
agriyakhetarpal: Thanks for the pointer! I'll keep that in mind and add a Ruff rule for this in another PR.

pyodide_build/cli/skeleton.py
agriyakhetarpal: Makes sense to me, sure! Addressed in 4e099e39d78fb74fb48b883c1df581c51f99410d.

pyodide_build/xbuildenv.py
agriyakhetarpal: Addressed in de4e3d70c5797ad010e72b9b694a20084ccdccfd.

PR #32, opened 2024-09-18
chore: do not use f-strings in logging statements

agriyakhetarpal:
## Description

Thanks to the tip in https://github.com/pyodide/pyodide-build/pull/30#discussion_r1764699168, I enabled a Ruff rule to avoid using f-strings in future logging statements we add: https://docs.astral.sh/ruff/rules/logging-f-string/, and converted the previous ones to comply with the rule. It's not necessarily required, but it's good to enable best practices.

> [!IMPORTANT]  
> This PR should be merged with a merge commit, not squashed.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Enable Ruff G004 to enforce logging best practices
+4c3043200ca15ade613fa3e3f8ca84bbe3522f21
--- a/pyodide_build/_py_compile.py
+++ b/pyodide_build/_py_compile.py
@@ -114,7 +114,7 @@ def _compile(
     output_name = output_path.name
 
     with set_log_level(logger, verbose):
-        logger.debug(f"Running py-compile on {input_path} to {output_path}")
+        logger.debug("Running py-compile on %s to %s", input_path, output_path)
 
         if compression_level > 0:
             compression = zipfile.ZIP_DEFLATED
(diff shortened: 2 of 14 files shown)


hoodmane (approved):
Thanks for doing this chore @agriyakhetarpal!

agriyakhetarpal:
Thanka for reviewing it!

agriyakhetarpal:
Thanks for reviewing it!

.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# Enable Ruff G004 to enforce logging best practices
+4c3043200ca15ade613fa3e3f8ca84bbe3522f21
```
ryanking13, line 2: What's this file for?

agriyakhetarpal: When we make formatting/style changes in the code that are not useful for the `git blame`, we can ignore the commits (revisions) in which the changes were made in the relevant files to keep their history clean.

More on this here: https://graphite.dev/guides/git-blame-ignore-revs

https://github.com/orgs/community/discussions/5033

agriyakhetarpal:
Also, oops, I was using GitHub through my phone and I squash merged it instead of creating a merge commit 🤦🏻 Let me fix this in a new PR.

agriyakhetarpal:
Fixed this in the `main` branch because we would have had to remove the squashed commit anyway. I had to force push; sorry for the inconvenience!

.git-blame-ignore-revs
agriyakhetarpal: Now that I fixed it – if you check any of the files, say, [`main`/pyodide_build/buildall.py (blame)](https://github.com/pyodide/pyodide-build/blame/main/pyodide_build/buildall.py), it wouldn't show the formatting change in the history (by default).

PR #36, opened 2024-09-25
MAINT Subclass recipe builders

ryanking13:
Subclasses RecipeBuilder based on the package type. This is to make it easier to distinguish between behavior that varies by package type. No functional changes intended.

I'll update the `buildall.py` as a follow-up.

I also removed prettier from the pre-commit, which I don't think is very useful for pyodide-build.


pyodide_build/buildpkg.py
```diff
@@ -412,6 +375,105 @@ def _compile(
                 self.src_extract_dir, self.src_dist_dir, build_env, config_settings
             )
 
+    def _patch(self) -> None:
```
ryanking13, line 378: There is no change in `_patch` or other functions. It seems git diff is not smart enough to create a small diff.

agriyakhetarpal (approved):
Thanks, @ryanking13, looks like nice maintenance!

agriyakhetarpal:
Prettier was quite helpful for me when linting YAML files for GitHub Actions, but I have no objections to removing it. We can use the VS Code extension for that.

ryanking13:
Thanks for the review!

> Prettier was quite helpful for me when linting YAML files for GitHub Actions, but I have no objections to removing it. We can use the VS Code extension for that.

Yeah, it is. But [mirrors-prettier](https://github.com/pre-commit/mirrors-prettier) is deprecated, so I think we should drop it. There aren't much yaml files in this repository so I think it is okay

PR #40, opened 2024-10-11
Insert -Zbuild-std when building rust packages

hoodmane:
Otherwise we are linking a copy of the standard library built with some particular version of emscripten selected by the rust compiler that does not match the version we are using. This can lead to ABI mismatch.
https://github.com/rust-lang/rust/issues/131467

cc @messense @davidhewitt perhaps it would be good for maturin to also insert this flag when targeting emscripten.

- [ ] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry

### DIFF
--- a/pyodide_build/pywasmcross.py
+++ b/pyodide_build/pywasmcross.py
@@ -525,6 +525,9 @@ def handle_command_generate_args(  # noqa: C901
     elif cmd == "strip":
         line[0] = "emstrip"
         return line
+    elif cmd == "cargo":
+        line.insert(1, "-Zbuild-std")
+        return line
     else:
         return line
 
--- a/pyodide_build/build_env.py
+++ b/pyodide_build/build_env.py
@@ -21,6 +21,7 @@ from .recipe import load_all_recipes
 RUST_BUILD_PRELUDE = """
 rustup toolchain install ${RUST_TOOLCHAIN} && rustup default ${RUST_TOOLCHAIN}
 rustup target add wasm32-unknown-emscripten --toolchain ${RUST_TOOLCHAIN}
+rustup component add rust-src --toolchain ${RUST_TOOLCHAIN}
 """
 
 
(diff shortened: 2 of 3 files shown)


hoodmane:
@juntyr says:

> When building the Rust crates, you could pass `--sysroot <PATH>` to a prebuilt sysroot. To each crate, it would look again just normal (no `build-std` anymore).
> 
> I'm not sure how to prepare the sysroot itself, but there is a crate https://crates.io/crates/cargo-sysroot that supports building with custom sysroots, so it might be useful to look at. 

This would have the advantage of not requiring nightly for this feature, but it has the disadvantage that none of the tooling looks very maintained...

pyodide_build/pywasmcross.py
```diff
@@ -525,6 +525,9 @@ def handle_command_generate_args(  # noqa: C901
     elif cmd == "strip":
         line[0] = "emstrip"
         return line
+    elif cmd == "cargo":
+        line.insert(1, "-Zbuild-std")
+        return line
```
ryanking13, line 530: Let's add a test for this (test_pywasmcross.py)

ryanking13 (approved):
Thanks! Well this is very tricky problem.

davidhewitt:
There is also https://github.com/RalfJung/rustc-build-sysroot which is maintained by a Rust project member but doesn't have a CLI; you'd need to wrap it.

I think both those tools do still need a nightly compiler to generate the sysroot though.

ryanking13:
@davidhewitt Thanks for the info! `rustc-build-sysroot` looks very intersesting. I am not sure patching rust packages to include that crate is the right way to go for us, though, as we are not package maintainers. 

Anyway, for now, we are having an issue that the `build-std` flag is incompatible with the `cargo vendor`, and  `rustc-build-sysroot` can be an alternative option.

hoodmane:
I think for now we can bump to a rust nightly that contains https://github.com/rust-lang/rust/pull/131533 and that will probably fix the trouble without `-Zbuild-std`. But in general, we have a pretty serious problem considering that we both must pass `-Zbuild-std` in order for compilation to be sound but also cannot pass it in order to compile packages with frozen crates. But we can add an option to suppress it and pass it for all other packages. And then just pray.

PR #42, opened 2024-10-16
Remove `-Zlink-native-libraries=no` from `RUSTFLAGS`

hoodmane:
The upstream bug that made this necessary seems to have been resolved: 
https://github.com/emscripten-core/emscripten/issues/16680

### DIFF
--- a/pyodide_build/config.py
+++ b/pyodide_build/config.py
@@ -195,7 +195,7 @@ DEFAULT_CONFIG: dict[str, str] = {
     "pyo3_config_file": str(TOOLS_DIR / "pyo3_config.ini"),
     "meson_cross_file": str(TOOLS_DIR / "emscripten.meson.cross"),
     # Rust-specific configuration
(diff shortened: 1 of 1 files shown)


ryanking13 (approved):
Nice! Thanks @hoodmane.

Could you please update the changelog? I think mentioning the Emscripten version that resolved this issue would be useful... but I am not sure which version of Emscripten actually fixed it.

hoodmane:
I have no idea either, it will require a bisect...

hoodmane:
Also, ironically now `-lc` works in side modules and is broken in the main module with `-sWASM_BIGINT`:
https://github.com/rust-lang/rust/pull/131736
https://github.com/emscripten-core/emscripten/pull/22765

ryanking13:
> I have no idea either, it will require a bisect...

No problem. We haven't updated the Emscripten version in Pyodide for a while, so it should be safe to conclude that it works in Pyodide >= 0.26

PR #46, opened 2024-10-31
Properly parse the content disposition filename

juntyr:
If the content disposition header filename uses quotes, `pyodide-build` includes the quotes in the literal filename, which e.g. confuses `shutil.unpack_archive` in determining the file type.

This PR ~~includes a simple fix to check and strip the quotes~~ uses the `email.message.Message` parser to properly extract the filename (see https://stackoverflow.com/a/78073510).

~~The PR should be squashed (the first commit is overriden by the second one).~~

### DIFF
--- a/pyodide_build/buildpkg.py
+++ b/pyodide_build/buildpkg.py
@@ -7,13 +7,13 @@ Builds a Pyodide package.
 import fnmatch
 import http.client
 import os
-import re
 import shutil
 import subprocess
 import sys
 import warnings
 from collections.abc import Iterator
 from datetime import datetime
+from email.message import Message
 from pathlib import Path
 from typing import Any, cast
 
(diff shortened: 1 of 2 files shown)


pyodide_build/buildpkg.py
```diff
@@ -330,6 +330,8 @@ def _download_and_extract(self) -> None:
             )
             if filenames:
                 tarballname = filenames[0]
+                if tarballname[0] == '"' and tarballname[-1] == '"':
+                    tarballname = tarballname[1:-1]
```
ryanking13, line 334: How about

tarballname.strip('"')?

juntyr:
> Add a test please?

What kind of test would you like? There's one static lib package in my fork that triggers the problem, but of course their server could stop wrapping the filename in quotes altogether. We could also extract the header parsing into a helper function and unit test it - if so I could also expected failure tests for other cases of the header that aren't handled yet (until someone else needs them).

Alternatively, is there some library that could do the header parsing for us so we don't need these half-baked fixes?

juntyr:
> Add a test please?

@hoodmane We're now using the `email.message.Message` parser and no custom logic anymore. Would you still like a test? I checked that all cases in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition are properly handled by this standard library package.

hoodmane:
It would still be good to have a test, particularly because it sounds like it will be quite simple. If there is some reason that adding a test would be very hard it can be omitted but that doesn't seem to me to be the case here.

hoodmane (approved):
Thanks!

hoodmane:
Just have to fix pre-commit...

juntyr:
@hoodmane This should now be good to go :)

hoodmane:
Thanks @juntyr!

PR #49, opened 2024-11-06
Apply PIP_CONSTRAINT correctly

hoodmane:


agriyakhetarpal (approved):

agriyakhetarpal:
We could also do a 0.29.1 release with this change, please let me know if you'd like that!

hoodmane:
We don't need to release just to merge then I can use the pyodide/pyodide-build commit on main.

agriyakhetarpal:
Sure, just re-triggered the CI failure because I'm unsure if it was flaky or not. Please feel free to merge if it passes

agriyakhetarpal:
Looks like a genuine failure.

hoodmane:
Looks like a real failure to me.

agriyakhetarpal:
I'm unable to reproduce this locally on macOS, the test passes for me. It's unrelated to `pytest`.

ryanking13:
Wait, but https://github.com/pyodide/pyodide/pull/5132 worked without this change. Why is this change needed?

hoodmane:
I'm not entirely sure why the other fix worked, but notice that we do not pass `build_env` to `install_reqs` in any way. So if we don't set this as an enviornment variable, how would the pip install command know that we have the constraint?

ryanking13:
`PIP_CONSTRAINT` is a host environment variables, so it should already be stored in os.environ without having to pass it in from build_env.

hoodmane:
When I ran it locally I found `os.environ["PIP_CONSTRAINT"]` raised a `KeyError` at this spot.

ryanking13:
Hmm, that's weird... 🤔

hoodmane:
Can you try it out and see if you log `os.environ["PIP_CONSTRAINT"]` here? If it prints something for you maybe we can just close this.

ryanking13:
Oh, you are right. It is slightly more complicated than I thought though. So what happens is:

1. Building packages **in-tree** with invoking `make`:
 
- Makefile will read and execute Makefile.envs, so `PIP_CONSTRAINT` is set in host env.

2. Building packages **in-tree** with invoking `pyodide build-recipes`.

- Makefile.envs will not be executed, so PIP_CONSTRAINT is not set in host env, but only stored in build env

3. Building packages **out-of-tree**

- Makefile.envs will not be executed, so PIP_CONSTARINT is not set in host env, but only stored in build env

So the current change looks okay. The reason why https://github.com/pyodide/pyodide/pull/5132 worked is probably because we try again with [build env replace](https://github.com/pyodide/pyodide-build/blob/8cebf75a5b62c68ce802d337a9a8c38c78068d2f/pyodide_build/pypabuild.py#L169-L175) when installing requirement fails.

pyodide_build/pypabuild.py
```diff
@@ -153,6 +153,7 @@ def _build_in_isolated_env(
 
         # first install the build dependencies
         symlink_unisolated_packages(env)
+        os.environ["PIP_CONSTRAINT"] = build_env["PIP_CONSTRAINT"]
```
ryanking13, line 156: Let's add a comment that this is only for in-tree build for future readers.

For OOT build, a user needs to set PIP_CONSTRAINT if needed.

hoodmane:
I think this was closed by https://github.com/pyodide/pyodide-build/pull/135.

PR #52, opened 2024-11-07
Set EM_PKG_CONFIG_PATH to correctly configure pkg-config

### LINKED ISSUE #51: pkg-config does not find dependencies in library_install_prefix
### Problem

In my build scripts, `pkg-config` could not find the dependencies in `library_install_prefix` as expected. It seems like pyodide-build tries to configure `pkg-config` here:

https://github.com/pyodide/pyodide-build/blob/4805d73e37e8448b7937faeb663bd8e450125417/pyodide_build/buildpkg.py#L552-L554

But in a call to `emconfigure ./configure ...` emscripten will [override this variable](https://github.com/emscripten-core/emscripten/blob/47b9540aa4364d5c1bde36859f03473d1f926838/tools/building.py#L65-L66).

### Proposed solution

I think it is necessary to set `EM_PKG_CONFIG_PATH`. This can be done in addition to `PKG_CONFIG_LIBDIR`. I'm unsure if the existing `PKG_CONFIG_LIBDIR` has any observable effect.

swnf:
Closes #51

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `skip_emscripten_version_check` flag and SKIP_EMSCRIPTEN_VERSION_CHECK environment
   variable to skip emscripten version check.
   [#53](https://github.com/pyodide/pyodide-build/pull/53)
+- Set the `EM_PKG_CONFIG_PATH` environment variable used by emscripten/`pkg-config` to discover dependencies
+  [#52](https://github.com/pyodide/pyodide-build/pull/52)
+
+### Changed
+
+- Source tar files are now extracted with python's [data filter](https://docs.python.org/3/library/tarfile.html#tarfile.data_filter)
+  [#52](https://github.com/pyodide/pyodide-build/pull/52)
 
 ## [0.29.0] - 2024/09/19
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [0.29.1]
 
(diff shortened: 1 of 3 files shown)


hoodmane (approved):
Makes sense to me, thanks!

ryanking13:
Looks okay to me, but could you please test it with your PR to verify (https://github.com/pyodide/pyodide/pull/5154)?

You can change the pyodide-build version used in our PR (https://pyodide.org/en/latest/development/maintainers.html#updating-pyodide-build)

swnf:
I pushed a new commit to pyodide/pyodide#5154 that uses the updated pyodide-build

ryanking13 (approved):
Thanks. Could you please update [the changelog](https://github.com/pyodide/pyodide-build/blob/main/CHANGELOG.md)?

swnf:
@ryanking13 I don't think this PR is ready to be merged. The [second commit](https://github.com/pyodide/pyodide-build/pull/52/commits/ec4f44018d6257044c57276f2721e536d284a1b2) I pushed to fix random pipeline failures might be related to a [scipy test failure](https://app.circleci.com/pipelines/github/pyodide/pyodide/8114/workflows/91f806ed-13e0-406d-9260-f83a04d77177/jobs/118120) (pipeline failed three times). I will try to investigate this issue and add a changelog after that.

swnf:
The pipeline failure seems to be caused by two issues with scipy:
 - The file `libopenblas.so` is not included in the wheel
 - The openblas build script patches `dist/libopenblas.so` to include libf2c (static library). But scipy explicitly uses `$WASM_LIBRARY_DIR/lib/libopenblas.so` which is not patched. This causes the libf2c symbols (like `s_cmp`) to be missing.

I've pushed [a64975df](https://github.com/pyodide/pyodide/pull/5154/commits/a64975df2870d02f1c0a114ac9570d839194379d) to my PR pyodide/pyodide#5154 to fix this. The pipeline is now [green](https://app.circleci.com/pipelines/github/pyodide/pyodide/8139/workflows/eecc43f8-3725-4154-ab06-a46588341c83). I have no clue why the build worked before or what exactly broke it.

This PR should now be ready to be merged. I've rebased it on `main` and updated the changelog.

PR #53, opened 2024-11-07
Add option to skip emscripten version check

ryanking13:
Adds an option to skip emscripten version check. A user can either set `SKIP_EMSCRIPTEN_VERSION_CHECK` env variable, or `skip_emscripten_version_check` in the toml file.

I think it is useful in following scenarios:

- When we update the Emscripten version in Pyodide but it is not released yet, so a user wants to build a package against newer emscripten version.
- When a user needs a newer version of Emscripten to make their package work. But the Emscripten version in Pyodide is not updated yet.

Of course, these scenarios will not work if two versions of Emscripten are incompatible. But I think it is useful to provide this option as not always

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Add `skip_emscripten_version_check` flag to skip emscripten version check.
+  [#53](https://github.com/pyodide/pyodide-build/pull/53)
+
 ## [0.29.0] - 2024/09/19
 
 ### Added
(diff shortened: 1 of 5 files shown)


agriyakhetarpal:
I think we should still check for "-git" in the version string in addition to this change.

ryanking13:
> I think we should still check for (and allow) "-git" in the version string in addition to this change.

Right. Let's fix it in this PR too.

agriyakhetarpal (approved):
Thanks, @ryanking13!

CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Add `skip_emscripten_version_check` flag to skip emscripten version check.
```
agriyakhetarpal, line 12:
```suggestion
- Add `skip_emscripten_version_check` flag and SKIP_EMSCRIPTEN_VERSION_CHECK environment
variable to skip emscripten version check.
```

Since we don't have docs for `pyodide-build` deployed yet, maybe this would be more explicit/helpful?

agriyakhetarpal (approved):

CHANGELOG.md
ryanking13: Sounds good.

ryanking13:
Thanks for the review!

PR #60, opened 2024-11-11
Use a more careful check in `remove_avoided_requirements`, and add PyWavelets recipe for testing

hoodmane:
Resolves https://github.com/pyodide/pyodide/issues/5187. In that issue, the build requires a package called `cmake-build-extension` but because we have `cmake` in the list of requirements to avoid we drop it and it breaks the build.

This logic was introduced in https://github.com/pyodide/pyodide/pull/2272 and did not come up in the review. Surprisingly this is the first problem it has caused. Let's see if this change breaks anything in Pyodide CI.

- [ ] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry
- [ ] Add / update tests


agriyakhetarpal:
As CMake is no longer an avoided requirement post #141, I think packages that use `cmake-build-extension` should work now, and this PR can be revisited.

hoodmane:
Could we explicitly add `oldest-supported-numpy` to the avoided requirements?

ryanking13:
> Could we explicitly add `oldest-supported-numpy` to the avoided requirements?

Yes, if it works, then I am okay with adding it as a workaround. I'll work on #68 later to handle it in a better way.

agriyakhetarpal:
Is there an offending package that requires `oldest-supported-numpy` that we can use as an integration test?

ryanking13:
> Is there an offending package that requires oldest-supported-numpy that we can use as an integration test?

Not in our integration test I think.

agriyakhetarpal:
> > Is there an offending package that requires oldest-supported-numpy that we can use as an integration test?
> 
> Not in our integration test I think.

Could you please clarify? As in, do you mean we don't t have a package like this, or that we don't want to add an integration test for such a package?

I think this should be good to go, but CI in the Pyodide repo has been failing since quite some time now, which makes things difficult to verify.

ryanking13:
> Could you please clarify? As in, do you mean we don't t have a package like this, or that we don't want to add an integration test for such a package?

"do you mean we don't t have a package like this" this one. I think it is good to add a package that depends on NumPy.

agriyakhetarpal:
Okay, I've added PyWavelets, as it depends on NumPy and Cython at build time. It has a small codebase and is fairly fast to build as a result. However, `oldest-supported-numpy` is deprecated for quite some time now ever since NumPy v2 came out, so it's difficult to find maintained packages that still depend on it.

ryanking13:
> oldest-supported-numpy is deprecated for quite some time now ever since NumPy v2 came out, so it's difficult to find maintained packages that still depend on it.

Makes sense. I think it is okay for now. Maybe we can make `AVOIDED_REQUIREMENTS` configurable by user, so we can help users fix it when they face errors without updating the pyodide-build version.

agriyakhetarpal:
> Maybe we can make `AVOIDED_REQUIREMENTS` configurable by user, so we can help users fix it when they face errors without updating the pyodide-build version.

That sounds like a good idea. Do you think we should leave that for a follow-up PR? If yes, @hoodmane, please feel free to merge. :)

ryanking13 (approved):

agriyakhetarpal (approved):
Okay, I think we should merge this now, as:

- it fixes the original error, and
- it adds PyWavelets as a recipe that depends on NumPy, and
- adds oldest-supported-numpy, and
- my PR #187 currently builds on top of this PR.

Thanks, @hoodmane and @ryanking13!

PR #62, opened 2024-11-24
Validate Python version when initializing xbuildenv

ryanking13:
This adds a check when initializing xbuildenv (== `init_environment`) to make sure the local Python version used to install the xbuildenv was not changed.

This prevents the following scenario:

1. One has a local Python version (3.11.3) and installs Pyodide xbuildenv (0.25.1).
2. Then, one changes the local Python version to (3.12.1) using pyenv.
3. Pyodide xbuildenv 0.25.1 is no longer compatible to the local Python version, so it should fail.

Related issue: https://github.com/pyodide/pyodide-build/issues/44

### DIFF
--- a/pyodide_build/xbuildenv.py
+++ b/pyodide_build/xbuildenv.py
@@ -423,7 +423,8 @@ class CrossBuildEnvManager:
             raise ValueError(
                 f"local Python version ({version_local}) does not match the Python version ({version_on_install}) "
                 "used to create the Pyodide cross-build environment. "
-                "Please reinstall the xbuildenv, by running `pyodide xbuildenv uninstall` and then `pyodide xbuildenv install`"
+                "Please switch back to the original Python version, "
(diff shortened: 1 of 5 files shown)


pyodide_build/xbuildenv.py
```diff
@@ -389,6 +397,35 @@ def uninstall_version(self, version: str) -> None:
...
+        if not version_file.exists():
+            raise ValueError("Python version marker file not found")
+
+        version_local = build_env.local_versions()["python"]
+        version_on_install = version_file.read_text().strip()
+        if version_on_install != version_local:
+            raise ValueError(
+                f"local Python version ({version_local}) does not match the Python version ({version_on_install}) "
+                "used to create the Pyodide cross-build environment. "
+                "Please reinstall the xbuildenv, by running `pyodide xbuildenv uninstall` and then `pyodide xbuildenv install`"
```
hoodmane, line 426: Or maybe switch back to the original Python version?

hoodmane (approved):
Thanks!

PR #68, opened 2024-12-02
[DRAFT] Use alternative index url when installing build dependencies

### LINKED ISSUE #43: Publish and install build dependencies to/from alternative package index
## Proposal

I propose hosting multiple versions of packages used as a build dependency (== numpy) in an alternative package index, and let pyodide-build use it when installing build dependencies.

### Why?

When building packages out-of-tree, each package may require a different version of Numpy. For now, if a package requires numpy as a build dependency, we just replace it with the numpy that is included in the Pyodide distribution, and hope it work.

For example, after we update the Numpy version in Pyodide distribution >= 2.0, some packages that use Numpy < 2.0 in the build step will fail to build because of the version mismatch.

### How?

I've been publishing packages built in [pyodide/pyodide-recipes](https://github.com/pyodide/pyodide-recipes) to Anaconda package registry: [Anaconda Pyodide org](https://anaconda.org/pyodide/dashboard). For now, these packages cannot be installed in the Pyodide runtime (CORS issue), but we can still use it during build time.

pyodide-build will point to this package index when installing build dependencies.

### Issues / Concerns

- How should we handle the in-tree cases?
- After unvendoring recipes, what should change?

### Overview

As-is:

```mermaid
sequenceDiagram
    Note over pyodide-build: Start building a package
    PyPI->>pyodide-build: Install native build dependency
    xbuildenv->>pyodide-build: Replace cross-build files
    Note over pyodide-build: Continue building package
```

To-be:

```mermaid
sequenceDiagram
    Note over pyodide-build: Start building a package
    Pyodide Package Index->>pyodide-build: Install cross-compiled build dependency
    Note over pyodide-build: Continue building package
```

ryanking13:
Close #43

Not ready to be reviewed, opened for visibility.


ryanking13:
This PR is ready to be reviewed now. I'll open a PR in pyodide/pyodide to test this.

hoodmane:
I guess I owe you a review on this one too.

PR #70, opened 2024-12-09
Fix is_link_cmd command to support response file

ryanking13:
Fix https://github.com/pyodide/pyodide/issues/5131

This fixes the `is_link_cmd` function which is used to detect the linker invocation.

If the command is too long, build systems like Ninja (or CMake, etc) use a response file to store the commands into a file. For instance, the command

```
emcc.py -o output.so input1.o input2.o ... inputN.o
```

becomes

```
emcc.py @output.so.rsp
```

where the file `output.so.rsp` stores all the commandline parameters.

Therefore, this PR fixes the `is_link_cmd` to detect the response file.

Note that I also removed the usage of `re` module, as I found that using string commands are much cheaper than compiling the regex.

- [ ] changelog

### DIFF
--- a/pyodide_build/pywasmcross.py
+++ b/pyodide_build/pywasmcross.py
@@ -83,11 +83,18 @@ def is_link_cmd(line: list[str]) -> bool:
     """
     Check if the command is a linker invocation.
     """
-    import re
-
-    SHAREDLIB_REGEX = re.compile(r"\.so(.\d+)*$")
     for arg in line:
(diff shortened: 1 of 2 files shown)


ryanking13:
This has some issue with meson response file (https://github.com/pyodide/pyodide/pull/5253#issuecomment-2541713573). Chaning to draft until I found a better way to integrate meson responsfile with pyodide-build.

PR #74, opened 2024-12-23
Fix handling cpython_module type recipe

ryanking13:
#36 had a bug not supporting `cpython_module` type recipes.


ryanking13:
It's a small bug fix, merging.

agriyakhetarpal:
Thanks for the catch!

PR #79, opened 2025-01-11
MAIN Subclass recipes more

ryanking13:
Related: #36

Continues to separate classes for each recipe type to make your code more structured, removing `if-else` conditions. NFC


agriyakhetarpal (approved):
Thanks, @ryanking13! I spent some time reviewing the changes and couldn't find any silent bugs with this refactor, so this should be ready to go in. Please feel free to merge.

ryanking13:
Thanks for the review!

hoodmane:
This breaks downstream `test_cpp_exceptions`. It seems that instead of `/usr/lib/cpp-exceptions-test-catch.so`, we find the shared library at `/lib/cpp-exceptions-test-catch.so`. I don't really understand what the bug is.

ryanking13:
Huh, interesting. Thanks. Let me check the failure.

ryanking13:
Found the reason. Thanks for letting me know! https://github.com/pyodide/pyodide-build/pull/110

PR #81, opened 2025-01-13
Build with wasm exception handling

hoodmane:
For now, we need to use a custom rust toolchain for this. Apparently all alternatives to this don't work. -Zbuild-std doesn't work with panic=abort (https://github.com/rust-lang/cargo/issues/7359)
and my attempts to use a custom sysroot with either https://github.com/RalfJung/rustc-build-sysroot/ or https://github.com/DianaNites/cargo-sysroot/  seem to hit the same problem as with `-Zbuild-std`. Thus, I think the only reasonable way to go is to build the sysroot from the rust source directory. Perhaps we can eventually approach this by copying the `lib/rustlib/wasm32-unknown-emscripten/lib/` folder out of the build of the rust compiler on top of a nightly install of the compiler. For now, there is the additional problem that we need this patch to fix unwind=abort:
https://github.com/rust-lang/rust/pull/135450

I got my copy of the rust compiler by checking out this commit: https://github.com/hoodmane/rust/commit/052ba16e2c1da50c80042cc7291b42e105a1e855 two commits ahead of the rust main branch and running:
```
./x build --stage 2 --target x86_64-unknown-linux-gnu,wasm32-unknown-emscripten
```


ryanking13:
Thanks for dealing with rust issues. I think we need some more discussion on how to support custom Rust toolchains.

Instead of hard-coding the values in the code, it would be nice to make this configurable, using `pyproject.toml` and `pyodide config`

Also, adding a new subcommand that can install some build tools, including rust toolchain can be useful IMO. There was an open issue about it (https://github.com/pyodide/pyodide/issues/3287).

hoodmane:
This also only works for x86-linux. I think what I have to do is build the rust compiler from source, then upload the `emscripten` static libraries folder, then use `rustup` to install a nightly, then manually add our build-from-source static libraries.

hoodmane:
But for that to work, we have to wait first for a nightly with https://github.com/rust-lang/rust/pull/135450.

ryanking13:
Another question is, when can we remove the custom rust toolchain and use the upstream one? I am okay with maintaining the rust toolchain and custom sysroot for a while, but I don't think we can do that forever. So it would be nice to keep track of the blockers.

Could you please open an issue that lists the blockers and their progress that needs to be resolved for the wasm exception handling?

hoodmane:
The plan is that we roll out wasm exception handling, check that there are no show stoppers, and then rust switches to using it by default after giving other people a chance to comment if it would cause them trouble.

hoodmane:
Once Rust turns it on by default, we can in principle use the normal Rust toolchain, with the one concern being that it could have abi mismatches due to being compiled with a different Emscripten version than we link with.

See the relevant MCP here:
https://github.com/rust-lang/compiler-team/issues/801

hoodmane:
I'll open tracking issues for the left over problems when we merge this.

hoodmane:
Also, @ryanking13 presumably we need to support both the `pyodide_2025_0_wasm32` abi and the `pyodide_2024_0_wasm32` at the same time? So I'll have to rework this a bit before merging to ensure this. Then we can merge this when the downstream pyodide/pyodide PR is ready and make a release and then all the externally built packages can be rebuilt and then they can be reenabled.

hoodmane:
Though I don't know if we want to make the Emscripten updates at the same time or just have a `2025_0` ABI with just wasm eh and a `2025_1` ABI with wasm eh plus Emscripten updates. Then the released ABI will use `2025_1`.

hoodmane:
Currently the only extra commit I need on top of the Rust main branch to make the appropriate toolchain is this:
https://github.com/hoodmane/rust/commit/81389fdc9d95a5a3f89e00239f6364d4f62d2d22

ryanking13:
> Though I don't know if we want to make the Emscripten updates at the same time or just have a 2025_0 ABI with just wasm eh and a 2025_1 ABI with wasm eh plus Emscripten updates. Then the released ABI will use 2025_1.

Maybe let's do both updates and call it `2025_0`.

pyodide_build/buildall.py
```diff
@@ -649,22 +650,18 @@ def run(self, n_jobs: int, already_built: set[str]) -> None:
...
-    rust_toolchain = build_env.get_build_flag("RUST_TOOLCHAIN")
+    toolchain_version = build_env.get_build_flag("RUST_TOOLCHAIN")
+    pyodide_root = get_pyodide_root()
+    toolchain_path = pyodide_root / ".rust-toolchain" / toolchain_version
+    if not toolchain_path.exists():
+        download_and_unpack_archive(
+            f"http://pyodide-cache.s3-website-us-east-1.amazonaws.com/rustc/{toolchain_version}.tar.bz2",
+            toolchain_path.parent,
+            "rust toolchain",
+        )
```
ryanking13, line 661: Lets add a configuration variable for this, and probably make it opt-in to download the toolchain from the custom URL.

For instance:

```python
rust_toolchain: "nightly-2024-01-29"
rust_toolchain_url_base: "" 
```

```python
toolchain_version = build_env.get_build_flag("RUST_TOOLCHAIN")
toolchain_url = build_env.get_build_flag("RUST_TOOLCHAIN_URL_BASE")

if toolchain_url:
    # if custom toolchain URL is given, download from it
else:
    # otherwise, download the standard toolchain
```

so we can,

1) change the URL without updating the pyodide-build version
2) make this feature optional, so people can still build packages against older pyodide versions easily

pyodide_build/pywasmcross.py
```diff
@@ -118,6 +118,9 @@ def replay_genargs_handle_dashl(arg: str, used_libs: set[str]) -> str | None:
     if arg == "-lgfortran":
         return None
 
+    if arg in ["-lfreetype", "-lpng"]:
+        arg += "-wasm-sjlj"
```
hoodmane, line 122: @ryanking13 maybe we can gate this on abi >= 2025_0? Or what do you think is the best way of doing this only when building against Pyodide 0.28 or later?

hoodmane: Or perhaps we should add a separate config variable for `WASM_EXCEPTIONS`.

pyodide_build/buildall.py
ryanking13: > maybe we can check for ABI 2025_0 or later? Because if you opt out you won't get an ABI compatible wheel.

Makes sense. Otherwise, if you think branching out using the ABI version will make the code too complex and increase the maintenance burden, I am okay with dropping support for Pyodide <= 0.27.X. We have [`min_pyodide_build_version` and `max_pyodide_build_version`](https://github.com/pyodide/pyodide/blob/main/pyodide-cross-build-environments.json) for that purpose.

pyodide_build/pywasmcross.py
ryanking13: I am okay with dropping support for Pyodide <= 0.27 if introducing the ABI or Pyodide version condition makes things too complicated.

However, this code seems to be very limited, only working for freetype and png libraries. Is there any way to specify this flag in the package recipe, not in pywasmcross.py?

hoodmane: Well the link args are generated by the build backend, so I don't see how. Unless they come from pkg-config or something.

hoodmane: I can add a comment to look into cleaning this up.

ryanking13: Yeah, we can leave it as TODO for now, but I would like to clean this up later.

ryanking13: Yeah, I agree. Instead of just adding "clean this up", how about adding some context for future readers then? I think there are a bunch of hacky things in this file, which probably is now outdated. But there are not enough comments in the code about the context, so it is hard to figure out.

pyodide_build/config.py
```diff
@@ -228,7 +230,8 @@ def _get_make_environment_vars(self) -> Mapping[str, str]:
     "rustflags": "-C link-arg=-sSIDE_MODULE=2 -C link-arg=-sWASM_BIGINT -Z link-native-libraries=no",
     "cargo_build_target": "wasm32-unknown-emscripten",
     "cargo_target_wasm32_unknown_emscripten_linker": "emcc",
-    "rust_toolchain": "nightly-2024-01-29",
+    "rust_toolchain": "nightly-2025-01-15",
```
ryanking13, line 233: Let's mention the default rust toolchain update in the changelog.

pyodide_build/pywasmcross.py
```diff
@@ -77,6 +77,7 @@ class CrossCompileArgs(NamedTuple):
     target_install_dir: str = ""  # The path to the target Python installation
     pythoninclude: str = ""  # path to the cross-compiled Python include directory
     exports: Literal["whole_archive", "requested", "pyinit"] | list[str] = "pyinit"
+    abi: str = ""
```
ryanking13, line 80: Let's add a comment about this parameter.

pyodide_build/pywasmcross.py
```diff
@@ -118,6 +119,9 @@ def replay_genargs_handle_dashl(arg: str, used_libs: set[str]) -> str | None:
     if arg == "-lgfortran":
         return None
 
+    if abi > "2025" and arg in ["-lfreetype", "-lpng"]:
+        arg += "-wasm-sjlj"
```
ryanking13, line 123: Let's add a comment about the context of this if condition.

pyodide_build/recipe/graph_builder.py
```diff
@@ -647,26 +648,50 @@ def run(self, n_jobs: int, already_built: set[str]) -> None:
...
+
+    # Install target from url
+    result = _run(
+        ["rustup", "which", "--toolchain", rust_toolchain, "rustc"],
+        capture_output=True,
+        text=True,
+    )
+    toolchain_root = Path(result.stdout).parents[1]
+    rustlib = toolchain_root / "lib/rustlib"
+    install_token = rustlib / "wasm32-unknown-emscripten_install-url.txt"
```
ryanking13, line 687: Is this file something you added manually in your custom toolchain? If so, let's add a comment about it.

pyodide_build/pywasmcross.py
```diff
@@ -578,7 +582,11 @@ def handle_command_generate_args(  # noqa: C901
     # Better to fail at compile or link time.
     if is_link_cmd(line):
         new_args.append("-Wl,--fatal-warnings")
-        new_args.extend(build_args.ldflags.split())
+        for arg in build_args.ldflags.split():
+            if arg.startswith("-l"):
+                arg = replay_genargs_handle_dashl(arg, used_libs, build_args.abi)
```
ryanking13, line 587: Not a strong opinion, but as `build_args.ldflags` is passed by the user, I think the user is responsible for setting the proper flags.

ryanking13 (approved):
Thanks! Codewise looks okay to me. I left some minor comments.

pyodide_build/pywasmcross.py
hoodmane: I have mixed feelings. I guess we can't fix static/dynamic libs not built with the python build system anyways so it's not that valuable. But it's pretty weird and annoying that linking `-lpng` successfully links but will break with opaque errors at runtime and the fix is to link with `-lpng-wasm-sjlj`. Maybe we can add some logic to audit wheel or somewhere appropriate that detects files linked against wrong libraries and explains as best we can what went wrong. Or people can just ask. 

So yeah probably it makes sense to remove this.

pyodide_build/config.py
hoodmane: This change isn't really needed to make this work. Maybe I can revert it here and put it by itself in a separate pr as a follow-up.

pyodide_build/pywasmcross.py
ryanking13: >  Maybe we can add some logic to audit wheel or somewhere appropriate that detects files linked against wrong libraries and explains as best we can what went wrong.

Yeah, I think the problem is that we are not building libraries like `libpng` or `libfreetype` ourselves and use flags like `-s USE_LIBPNG=1` instead. So, there exist multiple variants of those libraries, and it isn't easy to know what to use without looking into the emscripten code.

Ideally, I think it is better to drop all emscripten ports and build libraries ourselves. I think this would fix the issue with the Emscripten version and library version being fixed, and it would also fix the variant issues like the wasm exception.

ryanking13: BTW, I noticed that the library suffices are changed in the recent version of Emscripten: https://github.com/emscripten-core/emscripten/blob/242af002f8dee0366e86002fec187d447dcd26e5/tools/ports/libpng.py#L12-L16

pyodide_build/pywasmcross.py
```diff
@@ -118,6 +121,13 @@ def replay_genargs_handle_dashl(arg: str, used_libs: set[str]) -> str | None:
     if arg == "-lgfortran":
         return None
 
+    # Some Emscripten libraries that use setjmp/longjmp.
+    # The Emscripten linker should automatically know to use these variants so
+    # this shouldn't be necessary.
+    # This suffix will need to change soon to `-legacysjlj`.
+    if abi > "2025" and arg in ["-lfreetype", "-lpng"]:
+        arg += "-wasm-sjlj"
```
ryanking13, line 129: How about checking the `abi` version more strictly, so that this code will not break when we update Emscripten version and update ABI version?

ryanking13 (approved):
Thanks! I left a minor comment about ABI break, otherwise looks good to me

pyodide_build/pywasmcross.py
hoodmane: I don't know of a more strict check that makes sense to me here. Are you saying because Emscripten changed the suffix to `"legacysjlj`? But ABI doesn't contain the Emscripten version. I think when we do that Emscripten update we update this too.

ryanking13: For example, if we update the Emscripten version that changes the suffix + change the ABI version to 2025_1, this code will not work on that version. So I was curious if we can use `abi == 2025_0` here.

ryanking13: Of course, otherwise we can set `max_pyodide_build_version` in pyodide-cross-build-environments.json instead, explicitly specifying that this version of pyodide-build will not be compatible for future Pyodide versions.

hoodmane: I'm hoping to bring in that Emscripten update before we actually release. If you're concerned, perhaps I could add a check on the Emscripten version?

ryanking13: > I'm hoping to bring in that Emscripten update before we actually release.

Oh, I see. This would work too.

PR #83, opened 2025-01-15
Move ensure_rust_toolchain to buildall

hoodmane:
Rather than installing the rust toolchain in each build step that uses rust, do it once in buildall.

Work split off from #81.

### DIFF
--- a/pyodide_build/build_env.py
+++ b/pyodide_build/build_env.py
@@ -19,8 +19,7 @@ from pyodide_build.config import ConfigManager
 from pyodide_build.recipe import load_all_recipes
 
 RUST_BUILD_PRELUDE = """
-rustup toolchain install ${RUST_TOOLCHAIN} && rustup default ${RUST_TOOLCHAIN}
-rustup target add wasm32-unknown-emscripten --toolchain ${RUST_TOOLCHAIN}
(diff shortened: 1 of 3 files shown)


pyodide_build/build_env.py
```diff
@@ -19,8 +19,7 @@
 from pyodide_build.recipe import load_all_recipes
 
 RUST_BUILD_PRELUDE = """
-rustup toolchain install ${RUST_TOOLCHAIN} && rustup default ${RUST_TOOLCHAIN}
-rustup target add wasm32-unknown-emscripten --toolchain ${RUST_TOOLCHAIN}
+rustup default ${RUST_TOOLCHAIN}
```
hoodmane, line 22: I suppose we could also do this in `_ensure_rust_toolchain()` and remove `RUST_BUILD_PRELUDE` entirely.

ryanking13: Then , will `pyodide build-recipes-no-deps` work?

ryanking13 (approved):
Thanks. Mostly looks good to me. But please check it works correctly with `pyodide build-recipes-no-deps` CLI.

Also you can add `[integration]` in the commit message to run integration test, which includes building cryptography package.

PR #85, opened 2025-01-15
Add a new configuration variable for setting default xbuildenv url

ryanking13:
This PR adds a new config variable: `default_cross_build_env_url.`

This variable is used to set the default xbuildenv URL, which will be downloaded when a user calls `pyodide xbuildenv install` without passing the version, or when a user calls `pyodide build` without explicitly installing the xbuildenv.

__When is this variable useful?__

For now, building packages using the nightly cross-build env, we need to explicitly install it from URL:

```
pyodide xbuildenv install --url http://example.com/xbuildenv/dev/xbuildenv.tar.bz2
```

For instance, in pyodide-recipes repository, this is set in [CI script](https://github.com/pyodide/pyodide-recipes/blob/eb15b15360a8b11bdf510935120fa42e9be2311c/.github/workflows/build.yml#L111).

However, I would like to put this value in the `pyproject.toml`, which, I think is more appropriate place to put build-related variables.

### DIFF
--- a/pyodide_build/config.py
+++ b/pyodide_build/config.py
@@ -34,7 +34,7 @@ class ConfigManager:
     def _load_default_config(self) -> Mapping[str, str]:
         return deepcopy(DEFAULT_CONFIG)
 
-    def _load_cross_build_envs() -> Mapping[str, str]:
+    def _load_cross_build_envs(self) -> Mapping[str, str]:
         """
         Load environment variables from the cross build environment.
         """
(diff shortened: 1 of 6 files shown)


pyodide_build/config.py
```diff
@@ -13,8 +14,81 @@
 
 
 class ConfigManager:
+    """
+    Configuration manager for pyodide-build.
+    This class works "before" installing the cross build environment.
+    So it does not have access to the variables that are retrieved from the cross build environment.
```
ryanking13, line 20: The new config variable needs to be accessed before downloading the cross-build env, so I subclassed the `ConfigManager` into `ConfigManager` and `CrossBuildEnvConfigManager`. The former can be used even before installing the cross build env.

hoodmane: Can we separate splitting these into its own PR?

hoodmane (approved):
Thanks @ryanking13. I think everything makes sense and it's a nice feature. I would be nice to split the config as a separate PR since that part of the diff is quite noisy.

pyodide_build/config.py
ryanking13: Thanks. Sounds reasonable. Let me split the PR

ryanking13:
Thanks for the review!

PR #86, opened 2025-01-15
Update rust_toolchain, drop link_native_libs=no and bigint flags

hoodmane:
We don't need these anymore. link_native_libs=no was fixed by:
https://github.com/rust-lang/libc/pull/4002
and wasm_bigint was fixed by:
https://github.com/rust-lang/rust/pull/131736

### DIFF
--- a/pyodide_build/config.py
+++ b/pyodide_build/config.py
@@ -204,10 +204,10 @@ DEFAULT_CONFIG: dict[str, str] = {
     "cmake_toolchain_file": str(TOOLS_DIR / "cmake/Modules/Platform/Emscripten.cmake"),
     "meson_cross_file": str(TOOLS_DIR / "emscripten.meson.cross"),
     # Rust-specific configuration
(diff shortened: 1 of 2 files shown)


pyodide_build/config.py
```diff
@@ -204,10 +204,10 @@ def to_env(self) -> dict[str, str]:
     "cmake_toolchain_file": str(TOOLS_DIR / "cmake/Modules/Platform/Emscripten.cmake"),
     "meson_cross_file": str(TOOLS_DIR / "emscripten.meson.cross"),
     # Rust-specific configuration
-    "rustflags": "-C link-arg=-sSIDE_MODULE=2 -C link-arg=-sWASM_BIGINT -Z link-native-libraries=no",
+    "rustflags": "-C link-arg=-sSIDE_MODULE=2",
```
hoodmane, line 207: Actually, I guess this will make building for existing versions annoying. I'll leave these defaults here and just set `RUSTFLAGS` in `Makefile.envs`.

PR #91, opened 2025-01-24
Split config manager into `ConfigManager` and `CrossBuildEnvConfigManager`

ryanking13:
Splitted out from #85 to reduce diff.


ryanking13:
Approved in #85 so I'll go ahead and merge.

hoodmane:
Thanks @ryanking13 !

PR #96, opened 2025-01-29
Save symlinks directory when building with recipe builder

hoodmane:
Instead of putting the symlinks directory in a temp directory and deleting after the build is finished, put it under the build directory. This makes reproducing compile errors much easier, and the `build` directory is already just scratch files.

### DIFF
--- a/pyodide_build/pypabuild.py
+++ b/pyodide_build/pypabuild.py
@@ -243,6 +243,23 @@ def make_command_wrapper_symlinks(symlink_dir: Path) -> dict[str, str]:
 
 
 @contextmanager
+def _get_symlink_dir(env: dict[str, str], build_dir: Path | None):
+    if build_dir:
+        # If we're running under build-recipes, leave the symlinks in
+        # the build directory. This helps with reproducing.
+        symlink_dir = build_dir / "symlinks"
+        shutil.rmtree(symlink_dir, ignore_errors=True)
+        symlink_dir.mkdir()
+        yield symlink_dir
+        return
+
+    # Running from "pyodide build". Put symlinks in a temporary directory.
+    # TODO: Add a debug option to save the symlinks.
+    with TemporaryDirectory() as symlink_dir_str:
+        yield Path(symlink_dir_str)
+
+
+@contextmanager
 def get_build_env(
     env: dict[str, str],
     *,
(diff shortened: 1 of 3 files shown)


pyodide_build/pypabuild.py
```diff
@@ -242,6 +242,23 @@ def make_command_wrapper_symlinks(symlink_dir: Path) -> dict[str, str]:
     return env
 
 
+@contextmanager
+def _get_symlink_dir(env: dict[str, str], build_dir: Path | None):
```
ryanking13, line 246: Could you please check if it works well with `--continue` flag? I think it will, but just in case.

pyodide_build/pypabuild.py
```diff
@@ -242,6 +242,23 @@ def make_command_wrapper_symlinks(symlink_dir: Path) -> dict[str, str]:
...
+    if build_dir:
+        # If we're running under build-recipes, leave the symlinks in
+        # the build directory. This helps with reproducing.
+        symlink_dir = build_dir / "symlinks"
```
ryanking13, line 250: Maybe let's make the name more explicit and unique, such as `.pyodide_symlinks`.

pyodide_build/pypabuild.py
```diff
@@ -242,6 +242,23 @@ def make_command_wrapper_symlinks(symlink_dir: Path) -> dict[str, str]:
...
+        shutil.rmtree(symlink_dir, ignore_errors=True)
+        symlink_dir.mkdir()
+        yield symlink_dir
+        return
```
ryanking13, line 254: Considering that this function deletes and recreates the build directory. The name `_get_symlink_dir` is somehow misleading to me. How about renaming the function such as `_create_symlink_dir`?

ryanking13 (approved):
Thanks! I agree that keeping the symlinks is often quite useful. I left some comments otherwise looks good to me.

pyodide_build/pypabuild.py
hoodmane: I tried using `--continue` on failures in a few different places and it seems to work.

pyodide_build/pypabuild.py
hoodmane: Well we delete and recreate this build directory, the only things ever in it are:
1. the source folder
2. the source archive
3. now this symlink folder

I don't really see any reason to put a `.` in the name. But maybe I'll call it `pywasmcross_symlinks`.

Issue #87, opened 2025-01-16
Add `requirement.constraints` key in meta.yaml?


ryanking13:
For now, we are using a global `constraints.txt` file for building packages, but as the number of recipes grow, sometimes applying the same constraints to all packages doesn't fit.

For instance, https://github.com/pyodide/pyodide-recipes/pull/64#issuecomment-2595373923 had issue with the pined maturin version and the version that the package requires.

So, I think it would be useful to have a key in the `meta.yaml` that specifies the build constraints for the package:

```yaml
requirements:
  constraint:
    - maturin < 1.8  
```

and apply it only for that package.

cc: @chrysn

PR #97, opened 2025-02-01
Add `requirement.constraint` key to support package-wise constraints

ryanking13:
Resolve: #87

This PR adds a new key in meta.yaml file: `requirement.constraint`.

In the build time, pyodide-build will generate a new constraints.txt file in the build directory, and update the PIP_CONSTRAINT env variable to point that file. The new constarints.txt file contains `host_constraints_file + constraints defined in the recipe`.

- [x] changelog

### DIFF
--- a/pyodide_build/recipe/builder.py
+++ b/pyodide_build/recipe/builder.py
@@ -367,6 +367,7 @@ class RecipeBuilder:
             ldflags=ldflags,
             target_install_dir=self.build_args.target_install_dir,
             exports=self.build_metadata.exports,
+            build_dir=self.build_dir,
         )
         config_settings = pypabuild.parse_backend_flags(
             self.build_metadata.backend_flags
--- a/pyodide_build/recipe/builder.py
+++ b/pyodide_build/recipe/builder.py
@@ -338,9 +338,10 @@ class RecipeBuilder:
         shutil.move(self.build_dir / extract_dir_name, self.src_extract_dir)
         self.src_dist_dir.mkdir(parents=True, exist_ok=True)
 
-    def _override_constraints(self) -> str:
+    def _create_constraints_file(self) -> str:
         """
-        Override global constraints (PIP_CONSTRAINT) with constraints specific to this package.
+        Creates a pip constraints file by concatenating global constraints (PIP_CONSTRAINT)
+        with constraints specific to this package.
 
         returns the path to the new constraints file.
         """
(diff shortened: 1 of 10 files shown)


pyodide_build/recipe/builder.py
```diff
@@ -337,6 +338,33 @@ def _download_and_extract(self) -> None:
         shutil.move(self.build_dir / extract_dir_name, self.src_extract_dir)
         self.src_dist_dir.mkdir(parents=True, exist_ok=True)
 
+    def _override_constraints(self) -> str:
```
hoodmane, line 341: Maybe let's just call it `_get_constraints()`?

hoodmane (approved):
Thanks it looks very useful. It might be nice to add automation to override rust constraints later too.

pyodide_build/tests/recipe/_test_recipes/pkg_test_constraint/meta.yaml
```diff
@@ -0,0 +1,10 @@
+package:
+  name: pkg_test_constraint
+  version: "1.0.0"
+requirements:
+  constraint:
+    - numpy < 2.0
```
hoodmane, line 6: Maybe numpy and scipy are bad examples though since realistically this is broken I'd assume?

agriyakhetarpal:
I mostly agree with the functionality here, but I would like to suggest that we have simpler syntax – instead of having a `constraint:` field containing keys that are the same as  dependencies and are therefore redundant, could we add constraints to the `dependencies:` key itself? i.e., we can have two ways:

```yaml
requirements:
  constraint:
    - numpy <2
  executable:
    - rustup
  host:
    - numpy
  run:
    - numpy
    - scipy
```

and

```yaml
requirements:
  executable:
    - rustup
  host:
    - numpy <2
  run:
    - numpy >= 1.24
    - scipy >= 1.12.1, <1.14
```

The second option seems better, as it's closer to the conda-forge format. Also, this would allow adding constraints separately to the host and runtime dependencies (useful for build-time dependencies such as NumPy in this example), while `constraint:` seems to be applicable for both host and runtime dependencies simultaneously.

pyodide_build/recipe/builder.py
ryanking13: Thanks for the suggestion. I renamed it to `_create_constraints_file` as I thought that using the word `get` is confusing regarding that it creates a file internally.

ryanking13:
@agriyakhetarpal Thanks for the suggestion! I generally agree that it would be better to use the `build.host` key and `build.run` key to specify the constraints. It also aligns with the conda's meta.yaml specification.

But the problem is that currently, the `requirement` key in Pyodide's meta.yaml file does not align with the build dependency used by `pypa/build`. So it is a bit tricky to integrate it with pypa/build. So let's go as-is and think of a better way later.

PR #102, opened 2025-02-19
Add `FORCE_COLOR: 3` to GitHub Actions workflows

agriyakhetarpal:
## Description

This gets us better output for GitHub Actions activities; see https://force-color.org/. I feel all CI providers should enable this everywhere, but that's more of a radical opinion, perhaps 😄


agriyakhetarpal:
Merging this myself as this is a non-functional change (and an opinionated one as well, haha)

PR #103, opened 2025-02-19
Bump default Rust toolchain to `nightly-2025-02-01`

agriyakhetarpal:
## Description

`orjson` version 3.10.13 in #101 requires `rustc` 1.82 or newer, but we currently use 1.77-nightly ([see logs](https://github.com/pyodide/pyodide-build/actions/runs/13406461423/job/37447193337?pr=101)).

I have also performed the same update in https://github.com/pyodide/pyodide/pull/5249/ in parallel, so if that passes, we should be safe to merge this. I can also open a new PR with the submodule updated to check if that PR doesn't pass.

**Update, 23/02/2025:** all builds passed in pyodide/pyodide#5439.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+- The Rust toolchain version has been updated to `nightly-2025-01-18`.
+  [#103](https://github.com/pyodide/pyodide-build/pull/103)
+
 ## [0.29.3] - 2025/02/04
 
 ### Added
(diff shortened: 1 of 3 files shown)


ryanking13 (approved):
Thanks @agriyakhetarpal! I think this needs a changelog.

Also, could you run the integration test just in case?

agriyakhetarpal:
Thanks, Gyeongjae. https://github.com/pyodide/pyodide/pull/5249 doesn't pass, so I'll open a separate PR to test this change in Pyodide.

agriyakhetarpal:
`orjson`'s build passes with 2025-01-15 and fails with 2025-02-18, so I will bisect between both for a bit in the following few commits/CI runs.

agriyakhetarpal:
2025-02-01 passes for all packages locally, but I triggered CI in https://github.com/pyodide/pyodide/pull/5439 as an additional check.

agriyakhetarpal:
Going ahead here as suggested by @hoodmane in https://github.com/pyodide/pyodide/pull/5439#issuecomment-2676439259. Thanks for the review!

PR #105, opened 2025-02-20
Fix test with ANSI colour sequences + line breaks

agriyakhetarpal:
## Description

The `logger.success()` call in `make_package()` outputs colour schemes such as `\x1b[...]` (for example `\x1b[1;32m` means bold green text and `\x1b[0m` conveys resetting to default formatting), and the test isn't able to match the string because of these codes in between.

As it is just this one test in the entirety of our codebase that checks for the `/path/to/meta.yaml` string across multiple lines across the test suite, I decided to keep the logger as is instead of disabling the colours for successes.

This short PR fixes the test failure for me on local testing on a macOS machine. I'm still not sure how it has always passed in our CI tests. :D


ryanking13 (approved):
I think I missed this PR. Thanks!

PR #106, opened 2025-02-22
Add lower bound for typer

hoodmane:


ryanking13:
Could you please provide some context aoub this PR? If there is a breaking change in typer that doesn't match with Pyodide CLI, maybe we should apply this constraints to other packages as well?

hoodmane:
It's a lower bound though?

agriyakhetarpal:
I think Gyeongjae's suggestion is perhaps tied to the fact that adding the same lower bound to other `pyodide-cli` sub-packages can help ease dependency resolution when installing older versions of them. We should add lower bounds to more places wherever possible. :-)

hoodmane:
Makes sense. I think this is a small improvement that takes less work than working out and adding a lower bound to every single package in every single one of our repos. But that would be a good thing to do too. I just added this one because I ran into a crash here because my venv already had an old typer.

Issue #58, opened 2024-11-10
`pyodide xbuildenv install` won't work if installed in `uv`-managed virtual environments


agriyakhetarpal:
## Description

I stumbled into this issue where installing `pyodide-build` in a `uv`-managed venv and then running `pyodide xbuildenv install <...>` to set up a cross-build environment fails with the following logs, because we hardcode the `pip install <...>` command here:

https://github.com/pyodide/pyodide-build/blob/f78d587bd0e463619b6de13d6cfc8da3aaa9438b/pyodide_build/xbuildenv.py#L297-L312

 and `uv` removes the `pip` executable wrapper from its environments when activated so that a system `pip` isn't used in favour of `uv pip`:

<details>
<summary>Tap to show logs</summary>

```
❯ pyodide xbuildenv install 0.27.0a2
Downloading Pyodide cross-build environment from https://github.com/pyodide/pyodide/releases/download/0.27.0a2/xbuildenv-0.27.0a2.tar.bz2                    
Installing Pyodide cross-build environment                                                                                                                   
╭──────────────────────────────────────────────────────────── Traceback (most recent call last) ────────────────────────────────────────────────────────────╮
│ /Users/agriyakhetarpal/Desktop/numcodecs/venv/lib/python3.12/site-packages/pyodide_build/cli/xbuildenv.py:61 in _install                                  │
│                                                                                                                                                           │
│    58 │   if url:                                                                                                                                         │
│    59 │   │   manager.install(url=url, force_install=force_install)                                                                                       │
│    60 │   else:                                                                                                                                           │
│ ❱  61 │   │   manager.install(version=version, force_install=force_install)                                                                               │
│    62 │                                                                                                                                                   │
│    63 │   typer.echo(f"Pyodide cross-build environment installed at {path.resolve()}")                                                                    │
│    64                                                                                                                                                     │
│                                                                                                                                                           │
│ /Users/agriyakhetarpal/Desktop/numcodecs/venv/lib/python3.12/site-packages/pyodide_build/xbuildenv.py:221 in install                                      │
│                                                                                                                                                           │
│   218 │   │   except Exception as e:                                                                                                                      │
│   219 │   │   │   # if the installation failed, remove the downloaded directory                                                                           │
│   220 │   │   │   shutil.rmtree(download_path)                                                                                                            │
│ ❱ 221 │   │   │   raise e                                                                                                                                 │
│   222 │   │                                                                                                                                               │
[...]

Issue #54, opened 2024-11-07
Running via `pipx`/`uvx` isn't working fully


agriyakhetarpal:
## Description

We do support a `[project.entry-points."pipx.run"]` entry point, but it apparently doesn't work that well right now. I guess this is coming from our layered CLI.

## Steps to reproduce

The `pipx run pyodide build` command picks up the Pyodide sdist on PyPI:

<details>
<summary>Tap to show logs</summary>

<br>

```
PIP STDOUT
----------
Collecting pyodide
  Using cached pyodide-0.0.2.tar.gz (19 kB)
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'error'

PIP STDERR
----------
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [21 lines of output]
      Traceback (most recent call last):
        File "/Users/agriyakhetarpal/.local/pipx/shared/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
          main()
        File "/Users/agriyakhetarpal/.local/pipx/shared/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 335, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/Users/agriyakhetarpal/.local/pipx/shared/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 118, in get_requires_for_build_wheel
          return hook(config_settings)
                 ^^^^^^^^^^^^^^^^^^^^^
        File "/private/var/folders/b3/2bq1m1_50bs4c7305j8vxcqr0000gn/T/pip-build-env-dvq85nnc/overlay/lib/python3.12/site-packages/setuptools/build_meta.py", line 333, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=[])
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #107, opened 2025-02-22
Add some short installation and usage instructions

agriyakhetarpal:
## Description

This PR adds usage and installation instructions with `pip`, global installation and usage instructions with `pipx` and `uvx`, and some general recommendations.

This is mainly because I have these saved in my notes, which I have trouble finding at times – and the README could serve as a more permanent place for them until we have the documentation up and hosted.

Related to #54 and #58

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,43 @@ Tools for building Pyodide.
 See [http://github.com/pyodide/pyodide](http://github.com/pyodide/pyodide) for
 more information.
 
+## Quickstart
+
+### Usage
+
+#### Install it
+
+```bash
+pip install pyodide-build
+pyodide --help
+pyodide build --help
+```
+
+#### Install it globally
+
+```bash
+pipx install --include-deps pyodide-build
+pyodide --help
+pyodide build --help
+```
+
+or, alternatively
+
+```bash
+pipx install pyodide-cli
+pipx inject pyodide-cli pyodide-build
+pyodide --help
+pyodide build --help
+```
+
+### or run it
+
+- with `uv`/`uvx`
+
+```bash
+uvx --from pyodide-cli --with pyodide-build pyodide --help
+```
+
 ## License
 
 Pyodide uses the [Mozilla Public License Version
--- a/README.md
+++ b/README.md
@@ -9,6 +9,9 @@ more information.
 
 ### Usage
 
+> [!TIP]
+> Currently, installing `pyodide-build` is preferred to running it.
+
 #### Install it
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -37,6 +37,10 @@ pyodide --help
 pyodide build --help
 ```
 
+> [!NOTE]
+> If installing `pyodide-build` and its dependencies in virtual environments, we recommend
+using ones that are managed by `venv` or `virtualenv` at the moment; see #58.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,6 +5,50 @@ Tools for building Pyodide.
 See [http://github.com/pyodide/pyodide](http://github.com/pyodide/pyodide) for
 more information.
 
+## Quickstart
+
+### Usage
+
+> [!TIP]
+> Currently, installing `pyodide-build` is preferred to running it.
```
hoodmane, line 13: What does this mean? I usually don't have the question "should I install this or should I run it" because usually I have to install first in order to run?

README.md
```diff
@@ -5,6 +5,50 @@ Tools for building Pyodide.
...
+
+#### Install it
+
+```bash
+pip install pyodide-build
+pyodide --help
+pyodide build --help
```
hoodmane, line 20: Why do you include these?

README.md
```diff
@@ -5,6 +5,50 @@ Tools for building Pyodide.
...
+```bash
+pipx install pyodide-cli
+pipx inject pyodide-cli pyodide-build
+pyodide --help
+pyodide build --help
+```
+
+> [!NOTE]
+> If installing `pyodide-build` and its dependencies in virtual environments, we recommend
+using ones that are managed by `venv` or `virtualenv` at the moment; see #58.
```
hoodmane, line 42:
```suggestion
> [!NOTE]
> Currently `pyodide-build` does not work well in environments managed by uv, see #58.
```
I think this is the only thing in this PR that is appropriate for a readme. I think this adds a lot of noise, people can use pipx or uvx or venv or install pyodide-build globally or use whatever tools as they prefer. The readme isn't a tutorial on Python installer workflows.

README.md
agriyakhetarpal: Ah, I meant to convey that users are better off installing `pyodide-build` (globally or in a virtual environment) rather than trying to use `pipx` or `uvx` (instructions for which are provided for completeness). I'll try to reword this sentence.

README.md
agriyakhetarpal: Makes sense to me, thanks!

README.md
agriyakhetarpal: I can probably drop it, it's just meant to serve as a filler as we don't have docs hosted yet. But a CLI is usually self-documenting, so it's fine not to include this.

agriyakhetarpal:
One reason we could keep the global installation instructions is that they are used quite often in CI environments; `pipx` is used internally by GitHub actions to manage tooling and comes installed by default on GitHub runners.

Considering that `pyodide-build` doesn't provide a CLI of its own (which comes from `pyodide-cli`) and the custom `[pipx.run]` section can't help with this, it's not straightforward how to use it with such tools.

I can remove the unneeded bits, though. WDYT?

README.md
agriyakhetarpal: Addressed in b7708ff5707fda9d41d59cceb2bad1e52577f3fd.

README.md
agriyakhetarpal: I hope 3ec2ab2c63679c29a82b4fc6f7416bc7c29ad52b conveys the meaning better!

PR #108, opened 2025-02-22
Don't restore sys.platform in venv.py

hoodmane:
This is buggy when combined with pip 25. The commit when it was added mentions it was needed for `test_oot_build` to pass, but it seems to pass now.

Tested downstream:
https://github.com/pyodide/pyodide/pull/5455

- [x] Add a [CHANGELOG](https://github.com/pyodide/pyodide/blob/main/docs/project/changelog.md) entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- Fixed Pyodide venv `sys_platform` marker evaluation with pip >= 25.
+  [#108](https://github.com/pyodide/pyodide-build/pull/108)
+
 ## [0.29.3] - 2025/02/04
 
 ### Added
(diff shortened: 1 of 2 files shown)


hoodmane:
@ryanking13 have we thought about copying the `test_cmdline_runner` tests into `pyodide-build`? I guess that tests a pretty complicated mix of `pyodide` and `pyodide venv` so it's not clear where it's best to run it.

ryanking13:
> @ryanking13 have we thought about copying the test_cmdline_runner tests into pyodide-build? I guess that tests a pretty complicated mix of pyodide and pyodide venv so it's not clear where it's best to run it.

Yes, I thought about that when I was unvendoring pyodide-build. However, there was some question as to whether this should be included in pyodide-build, given that all other parts of pyodide-build do not rely on the Pyodide runtime, only pyodide venv.

Also, the `test_cmdline_runner` is very dependent on changes in the `python` CLI, which we didn't think was stable enough at the time.

Anyway, if we will keep evolve `pyodide venv` in the long run, I agree with moving the cmdline runner to pyodide-build.

PR #109, opened 2025-02-22
Respect `SOURCE_DATE_EPOCH` for better reproducibility of builds

agriyakhetarpal:
## Description

This PR add supports for the [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/specs/source-date-epoch/) environment variable when creating wheels from recipes to enable reproducible builds. The implementation follows the Python example from the https://reproducible-builds.org specification.

I can't say that this change makes builds completely reproducible, as we are responsible only for these utility functions and our usage of `pypa/build` as a build frontend – which doesn't guarantee reproducibility at all (see https://github.com/pypa/build/issues/385); that responsibility relies on the build backend, which we are agnostic to. It's a start towards more reproducibility, however.

I've modifies the `_make_whlfile` function to set the timestamps based on `SOURCE_DATE_EPOCH`. When `SOURCE_DATE_EPOCH` is provided, all files inside the wheel will have their timestamps set to the provided date. "January 1st 1980 00:00:00 UTC" (Unix timestamp `315532800`) is used as it is the minimum for ZIP file compatibility. A `reproducible_filter` has been added for tarfile creation that acts similarly to the "data" filter but with added `SOURCE_DATE_EPOCH` consideration.

### DIFF
--- a/pyodide_build/recipe/builder.py
+++ b/pyodide_build/recipe/builder.py
@@ -547,7 +547,7 @@ class RecipeBuilderPackage(RecipeBuilder):
     ) -> None:
         """Package a wheel
 
-        This unpacks the wheel, unvendors tests if necessary, runs and "build.post"
+        This unpacks the wheel, unvendors tests if necessary, and runs the "build.post"
         script, and then repacks the wheel.
 
         Parameters
--- a/pyodide_build/recipe/builder.py
+++ b/pyodide_build/recipe/builder.py
@@ -627,7 +627,24 @@ class RecipeBuilderPackage(RecipeBuilder):
                     )
                     if nmoved:
                         with chdir(self.src_dist_dir):
-                            shutil.make_archive(f"{self.name}-tests", "tar", test_dir)
+                            filetime = _get_source_epoch()
+                            make_archive_kwargs = {
+                                "root_dir": "tests",
+                                "owner": "root",
+                                "group": "root",
+                            }
+
+                            if "SOURCE_DATE_EPOCH" in os.environ:
+
+                                def _set_time(tarinfo):
+                                    tarinfo.mtime = filetime
+                                    return tarinfo
+
+                                make_archive_kwargs["filter"] = _set_time
+
+                            shutil.make_archive(
+                                f"{self.name}-tests", "tar", **make_archive_kwargs
+                            )
             finally:
                 shutil.rmtree(test_dir, ignore_errors=True)
 
--- a/pyodide_build/recipe/builder.py
+++ b/pyodide_build/recipe/builder.py
@@ -8,6 +8,7 @@ import os
 import shutil
 import subprocess
 import sys
+import time
 from collections.abc import Iterator
 from datetime import datetime
 from email.message import Message
@@ -44,6 +45,18 @@ from pyodide_build.recipe.bash_runner import (
     get_bash_runner,
 )
 from pyodide_build.recipe.spec import MetaConfig, _SourceSpec
+
+
+def _get_source_epoch() -> int:
+    """Get SOURCE_DATE_EPOCH from environment or fallback to current time.
+    Uses 315532800, i.e., 1980-01-01 00:00:00 UTC as minimum timestamp (as
+    this is the zipfile limit).
+    """
+    try:
+        source_epoch = int(os.environ.get("SOURCE_DATE_EPOCH", time.time()))
+        return max(315532800, source_epoch)
+    except ValueError:
+        return int(time.time())
 
 
 def _make_whlfile(
--- a/pyodide_build/recipe/builder.py
+++ b/pyodide_build/recipe/builder.py
@@ -8,6 +8,7 @@ import os
 import shutil
 import subprocess
 import sys
+import tarfile
 import time
 from collections.abc import Iterator
 from datetime import datetime
@@ -352,7 +353,7 @@ class RecipeBuilder:
                 _get_source_epoch() if "SOURCE_DATE_EPOCH" in os.environ else None
             )
 
-            def reproducible_filter(tarinfo):
+            def reproducible_filter(tarinfo: tarfile.TarInfo) -> tarfile.TarInfo:
                 """Filter that preserves permissions but normalizes ownership and optionally
                 timestamps. This is similar to the "data" filter but injects SOURCE_DATE_EPOCH."""
 
--- a/pyodide_build/recipe/builder.py
+++ b/pyodide_build/recipe/builder.py
@@ -345,11 +345,29 @@ class RecipeBuilder:
         # is too large for the chown() call. This behavior can lead to "Permission denied" errors
         # (missing x bit) or random strange `make` behavior (due to wrong mtime order) in the CI
         # pipeline.
-        shutil.unpack_archive(
-            tarballpath,
-            self.build_dir,
-            filter=None if tarballpath.suffix == ".zip" else "data",
-        )
+        if tarballpath.suffix == ".zip":
+            shutil.unpack_archive(tarballpath, self.build_dir, filter=None)
+        else:
+            filetime = (
+                _get_source_epoch() if "SOURCE_DATE_EPOCH" in os.environ else None
+            )
+
+            def reproducible_filter(tarinfo):
+                """Filter that preserves permissions but normalizes ownership and optionally
+                timestamps. This is similar to the "data" filter but injects SOURCE_DATE_EPOCH."""
+
+                tarinfo.uid = tarinfo.gid = 0
+                tarinfo.uname = tarinfo.gname = "root"
+
+                # set timestamp from SOURCE_DATE_EPOCH if available
+                if filetime is not None:
+                    tarinfo.mtime = filetime
+
+                return tarinfo
+
+            shutil.unpack_archive(
+                tarballpath, self.build_dir, filter=reproducible_filter
+            )
 
         extract_dir_name = self.source_metadata.extract_dir
         if extract_dir_name is None:
--- a/pyodide_build/recipe/builder.py
+++ b/pyodide_build/recipe/builder.py
@@ -62,6 +62,9 @@ def _get_source_epoch() -> int:
 def _make_whlfile(
     *args: Any, owner: int | None = None, group: int | None = None, **kwargs: Any
 ) -> str:
+    filetime = _get_source_epoch()
+    # gtime() ensures UTC
+    kwargs["date_time"] = time.gmtime(filetime)[:6]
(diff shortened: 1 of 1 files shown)


pyodide_build/recipe/builder.py
```diff
@@ -329,11 +346,31 @@ def _download_and_extract(self) -> None:
...
+        if tarballpath.suffix == ".zip":
+            shutil.unpack_archive(tarballpath, self.build_dir, filter=None)
+        else:
+            filetime = (
+                _get_source_epoch() if "SOURCE_DATE_EPOCH" in os.environ else None
+            )
+
+            def reproducible_filter(
+                tarinfo: tarfile.TarInfo, path: str | Path | None = None
+            ) -> tarfile.TarInfo:
```
agriyakhetarpal, line 358: I'm not a fan of this method being nested in another method. However, there are similar patterns in this file as well.

ryanking13 (approved):
Thanks @agriyakhetarpal! I think it is a good improvement. Also adding cross-ref for the previous discussion: https://github.com/pyodide/pyodide/discussions/4286

Could we have a test for the new behavior?

agriyakhetarpal:
Thanks for the review, @ryanking13 – I have added a bunch of tests, and also manually confirmed with the integration tests that the change works towards their reproducibility. This should resolve most of the conversation in https://github.com/pyodide/pyodide/discussions/4286. I'll leave a comment there about it, which should help, as issues/PRs cross-linked from discussions don't show up in the UI.

I had a couple of questions about the downstream integration of this feature based on that discussion:
- Should we apply `SOURCE_DATE_EPOCH` somewhere in our build process in the Pyodide repository, such as the Dockerfile? We could point to the timestamp when the Docker images are built from the Dockerfile using our manual GHA trigger. See: https://docs.docker.com/build/ci/github-actions/reproducible-builds/
- Should we have a test downstream that checks that the `pyodide-lock.json` will not be modified by repeatedly rebuilding packages, or is that mostly handled by the tests here?

hoodmane:
> Should we ...?

I'm not sure but both of these sound like good ideas to me. I know that there is a bit of no determinism in the binaries generated by emcc, not sure about shared libraries. Maybe we should open an issue on emscripten asking what the status of reproducible builds is.

hoodmane:
What feels to me like it would make the most sense would be to set the date based on when the latest commit occurred.

agriyakhetarpal:
I found https://github.com/emscripten-core/emscripten/issues/7714, though from the conversation I feel that it was closed as "won't fix" for an unrelated reason. I don't want to comment there asking for an update as it is quite old, but if the premise is that if LLVM guarantees reproducibility as per https://blog.llvm.org/2019/11/deterministic-builds-with-clang-and-lld.html, we should be fine with even level 1 determinism even if we can't have level 4 determinism right now.

One thing that could be nice is to switch the Pyodide runtime's versioning scheme to be based on Git tags. That could require a bit of effort because of all the moving parts everywhere, though (especially downstream).

pyodide_build/recipe/builder.py
agriyakhetarpal: Thanks, addressed in 4c55ea8d5816cbf7336b1f5b210a7a0d6b074d02 by moving to a private non-nested method.

pyodide_build/recipe/builder.py
```diff
@@ -308,6 +308,23 @@ def ignore(path: str, names: list[str]) -> list[str]:
 
         self.src_dist_dir.mkdir(parents=True, exist_ok=True)
 
+    def _reproducible_filter(
```
hoodmane, line 311: Doesn't this need to be a `@staticmethod` or `@classmethod`? And it has no self param so I assume it's currently crashing?

agriyakhetarpal: Yes, also it raises an error because I didn't get the arguments for the filter right. Let me correct that and trigger the integration tests again.

agriyakhetarpal: Never mind, they were correct here – fixed it.

agriyakhetarpal:
I'll mark it as a draft for now, as I notice that the `SOURCE_DATE_EPOCH` isn't being respected properly when I run `pyodide build-recipes`, but it works well with `pyodide build` on an individual source directory. I'll be attempting to generate a lockfile in the integration tests, comparing with which can ensure that the checksums aren't modified when rebuilding (as much as possible).

pyodide_build/recipe/builder.py
hoodmane: This is a method on `RecipeBuilder` right? So it should take a first argument called `self`? Or be marked `@staticmethod`? Or am I misreading it?

hoodmane: Github UI should say "outdated" then =(

hoodmane:
We filter environment variables. Is the problem that you haven't updated that logic to let `SOURCE_DATE_EPOCH` through?

agriyakhetarpal:
Ah, I didn't know that. Could you please share a pointer where we do so in the `pyodide build-recipes` phase?

agriyakhetarpal:
I've tried to pass it to the `BuildArgs` dataclass for a start, but I don't know if that's the right approach.

hoodmane:
You need to add it here:
https://github.com/pyodide/pyodide-build/blob/main/pyodide_build/config.py?plain=1#L156

Any environment variable not in `BUILD_KEY_TO_VAR` is filtered out.

hoodmane:
Make the recipe echo the environment variable in the build script.

agriyakhetarpal:
Yes, it makes it through: 

```
 26100K .......... .......... .......... .......... .......... 99% 16.5M 0s
 26150K .......... .......... .......... .......... .......... 99% 18.3M 0s
 26200K .......... .......... .......... .......... .......... 99% 14.0M 0s
 26250K .......... .......... .......... .......... .......... 99% 26.7M 0s
 26300K .......... .......... .......... .......... .......... 99% 15.7M 0s
 26350K .......... .......... .......... ...                  100% 12.2M=1.3s

2025-02-24 16:26:34 (19.6 MB/s) - ‘Python-3.12.7.tgz’ saved [27016272/27016272]

SOURCE_DATE_EPOCH: 1735689600
Successfully built /Users/agriyakhetarpal/Desktop/agriya-pyodide-build/integration_tests/recipes/pydoc_data/build/pydoc_data-1.0.0/dist/pydoc_data-1.0.0-py2.py3-none-any.whl
Unpacking wheel to                                                                                                                           
/Users/agriyakhetarpal/Desktop/agriya-pyodide-build/integration_tests/recipes/pydoc_data/build/pydoc_data-1.0.0/dist/pydoc_data-1.0.0-py2.py3
-none-any.whl                                                                                                                                
Unpacking to: /var/folders/b3/2bq1m1_50bs4c7305j8vxcqr0000gn/T/tmpek4du66r/pydoc_data-1.0.0...OK
Repacking wheel as /Users/agriyakhetarpal/Desktop/agriya-pyodide-build/integration_tests/recipes/pydoc_data/build/pydoc_data-1.0.0/dist/pydoc_data-1.0.0-py2.py3-none-any.whl...OK
[2025-02-24 16:26:35] Succeeded building package pydoc_data in 3.5 seconds.
```

That makes me think, maybe it is `pyodide-lock` that unpacks and repacks the wheels to get the metadata? I'm completely unfamiliar with its codebase at the moment, but could start reading it.

agriyakhetarpal:
That means we should be good to go for this PR, but we'll need to update `pyodide-lock` similarly, and get a new v0.1.0a9 release out for it to be used here. Thanks for the tips, @hoodmane!

agriyakhetarpal:
🤔 Building NumPy repeatedly outside of the lockfile generation displays different checksums every time for its wheels[^1]. Still, upon opening the wheel and the test archives, I see that `SOURCE_DATE_EPOCH` has been respected properly. We can't control Meson too much at all as a build frontend, but their motto is that non-reproducibility is a bug": https://mesonbuild.com/Reproducible-builds.html

However, both the `shasum -a 256 integration_tests/recipes/.libs/lib/libz.a` and `shasum -a 256 integration_tests/recipes/.libs/lib/libgeos.so` commands return the same checksum for `zlib` and `geos` respectively, indicating that their builds were reproducible indeed.

So, in general, I wonder if we're checking for the right things here by trying to validate checksums.

[^1]: This is possibly because NumPy's recipe has a `post:` section. Unvendoring the tests is reproducibly managed, but the post-script doesn't go through such code paths at the moment, I'll investigate.

agriyakhetarpal:
I think I'm running into the same problem I faced in https://github.com/pyodide/pyodide/pull/5031. Is there a way to add `--no-build-isolation` when building NumPy, or would we need to add support for that?

agriyakhetarpal:
TODO:

- [x] Reproducible builds for packages in the integration tests: `pydoc_data`, `geos`, and `zlib`
- [ ] Fix NumPy reproducibility issue by allowing it to not build under isolation
	- [ ] Extend `pypabuild.py` and our CLI to have a non-isolated option based on `pypa/build` code paths?
- [ ] Investigate the existence of post-script reproducibility issue
- [ ] Generate lockfile in integration tests and add a small bash/`jq` script as a test for reproducibility of sorts
	- [ ] Investigate why the lockfile generation step (possibly in `pyodide-lock`) modifies the checksums for packages

ryanking13:
> You need to add it here: https://github.com/pyodide/pyodide-build/blob/main/pyodide_build/config.py?plain=1#L156
> 
> Any environment variable not in `BUILD_KEY_TO_VAR` is filtered out.

I believe this is not true. If it does, it should be a bug. We passthrough all host env variables to the builder.

ryanking13:
> I think I'm running into the same problem I faced in https://github.com/pyodide/pyodide/pull/5031. Is there a way to add --no-build-isolation when building NumPy, or would we need to add support for that?

Thanks. Supporting `--no-build-isolation` would be a reasonable approach to fix this issue. But I also think it might a bit over-engineering just to solve the reproducibility issue. I think it is something that the downstream packages should handle, not the build system.

Anyway, we've got the request to support no isolation mode a few times, so it's probably worth discussing in a separate issue.

ryanking13:
> That makes me think, maybe it is pyodide-lock that unpacks and repacks the wheels to get the metadata? I'm completely unfamiliar with its codebase at the moment, but could start reading it.

Interesting. Probably [`wheel` CLI to unpack and repack the wheel](https://github.com/pyodide/pyodide-build/blob/aac59785e656a3b78e6906853e398201bf3ca3e4/pyodide_build/common.py#L289-L312) is changing the timestamp in the zipfile? It is not in pyodide-lock (but I think it will be moved to pyodide-lock in the future).

PR #110, opened 2025-02-23
Fix typo in shared library builder

ryanking13:
There was a typo that made the shared library installed in an unexpected directory. I guess we should validate this in pyodide-lock.


hoodmane:
> I guess we should validate this in pyodide-lock.

Indeed. Thanks for the quick fix!

PR #116, opened 2025-03-06
MAINT Unvendor tests at the lockfile creation step

ryanking13:
This PR changes when test codes are unvendored.

Before:

- it is unvendored right after the wheel file is built (== at the same step when `build.post` script runs).
- when generating a lockfile, we check if the test codes are unvendored and update the lockfile entry.

After:

- it is unvendored when we are creating a lockfile.

__Why?__

I would like to move the unvendoring logic to somewhere else like `pyodide-pack` or `pyodide-lock` (https://github.com/pyodide/pyodide-lock/issues/30, https://github.com/pyodide/pyodide-pack/issues/41), so package maintainers can use that feature when they are not using a recipe.

---

Related issues:

- https://github.com/pyodide/pyodide-lock/issues/30
- https://github.com/pyodide/pyodide-pack/issues/41


agriyakhetarpal (approved):
Thanks for making this change, @ryanking13! I agree that `pyodide build` as a build frontend should not alter the package to remove the test files; it should be the job of the recipe maintainers.

I think https://github.com/pyodide/pyodide-pack/issues/41 is the ideal location to move these tests to, as it will help that repository gain more mobility.

I would suggest that we don't merge/release this change until we release a version of `pyodide-pack` with that issue resolved. What do you think? I could spend some time implementing it, or step back if you are interested. :)

ryanking13:
> I think https://github.com/pyodide/pyodide-pack/issues/41 is the ideal location to move these tests unvendoring to, as it will help that repository gain more mobility.

Yeah, at first, I thought we should put it in `pyodide-lock`, but I realized that the reason why we made a separate repository for `pyodide-lock` was to make the lockfile definition as small as possible (which was a request from JupyterLite team).

So instead of putting complex logic in the `pyodide-lock`. Maybe let's keep the `pyodide-lock` as small as possible, and use other packages to do the job.

ryanking13:
> I would suggest that we don't merge/release this change until we release a version of pyodide-pack with that issue resolved. What do you think?

This change itself does not introduce any breaking changes (hence no changelog), so I think we can merge this as is. Then, I think you can start copying the logic to the pyodide-pack. After that, we can make `pyodide-pack` a dependency of pyodide-build.

> I could spend some time implementing it, or step back if you are interested. :)

I would be happy if you could work on this. TBH, no one has recently worked on pyodide-pack after Romans retirement, so you would be the best person who understands the codebase.

ryanking13:
I am not going to make a release though, as `"-wasm-sjlj"` part needs to be updated after we update the Emscripten version in Pyodide (https://github.com/pyodide/pyodide/pull/5332)

agriyakhetarpal:
> > I would suggest that we don't merge/release this change until we release a version of pyodide-pack with that issue resolved. What do you think?
> 
> This change itself does not introduce any breaking changes (hence no changelog), so I think we can merge this as is. Then, I think you can start copying the logic to the pyodide-pack. After that, we can make `pyodide-pack` a dependency of pyodide-build.
> 
> > I could spend some time implementing it, or step back if you are interested. :)
> 
> I would be happy if you could work on this. TBH, no one has recently worked on pyodide-pack after Romans retirement, so you would be the best person who understands the codebase.

Oh, I see. Yes, please feel free to merge this! I'll close https://github.com/pyodide/pyodide-lock/issues/30 for now, and proceed to work on `pyodide-pack` soon.

Issue #115, opened 2025-03-05
sys_platform != 'emscripten' in "dependencies" doesn't work


antocuni:
## 🐛 Bug

`sys_platform` has the wrong value when doing `pip install`. For example:

```
# pyproject.toml
[project]
dependencies = [
    "ziglang==0.13.0; sys_platform != 'emscripten'",
]
```

```
$ ./pyodide-venv/bin/pip install .
```

This will try to install ziglang, even if it shouldn't.

### To Reproduce

I got this problem when trying to install `spy` in a pyodide env. You can get it from this PR:
https://github.com/spylang/spy/pull/128

### Expected behavior

`pip` should not try to install `ziglang`.

Note that you can fix it "manually" by modifying `./pyodide-venv/bin/pip` in this way:

```diff
--- pip.old	2025-03-06 00:22:09.060759680 +0100
+++ pip	2025-03-06 00:22:17.167805760 +0100
@@ -41,7 +41,7 @@
 import sysconfig
 sysconfig._init_config_vars()
 del os.environ["_PYTHON_SYSCONFIGDATA_NAME"]
-sys.platform = orig_platform
+#sys.platform = orig_platform
 import re
 import sys
 from pip._internal.cli.main import main
```

### Environment

- Pyodide Version: 0.27.2

hoodmane:
What's your pip version @agriyakhetarpal ?

hoodmane:
In any case, the code to restore `sys.platform` was added in the python 3.11 update with a comment saying that it fixed a test failure in `test_cmdline_runner`. Since the change is a simplification and as far as I can tell the test doesn't fail anymore I think it's a simple enough change perhaps...

agriyakhetarpal:
> What's your pip version @agriyakhetarpal ? I think this is specifically a pip 25 problem.

Oh, I'm on 25.0.1 – the latest available

hoodmane:
If you have a moment could you test if you can reproduce with 25.0.0? Maybe it was a regression that got fixed?

hoodmane:
I don't see any relevant looking change log entries...

agriyakhetarpal:
I downgraded to `pip==25.0.0` but that apparently broke the virtual environment, as I get the usual `Emscripten does not support processes` error. Is there a convenient way to try out 25.0.0?

agriyakhetarpal:
[...]

PR #117, opened 2025-03-08
Make `pyodide venv` more compatible with standard `virtualenv` args + add tests

agriyakhetarpal:
## Description

This PR stems from https://github.com/pyodide/pyodide-build/issues/115#issuecomment-2706715796 where I noticed that it is currently not possible to extend `virtualenv` args in order to override the virtual environment creation behaviour.

I've added a subset of the arguments listed at https://virtualenv.pypa.io/en/stable/cli_interface.html. The environment variables are already supported as we don't have much to deal with them, this PR enables passing those arguments. A warning is raised on currently and to-be-always unsupported arguments. For example, `--no-pip` does not make sense as we need `pip` to be able to install packages, and thus we always need a seeded environment.

It _might_ be possible to add support for the `--copies` and `--symlink-app-data` flags as well, but I haven't looked into them as they seem to be a bit more complex than I expected. Also, I'm not sure if anyone will need the extra complexity.

Further steps that can be taken in the medium term are to create a [`virtualenv-pyodide` plugin](https://virtualenv.pypa.io/en/latest/extend.html) and register it with `virtualenv` so that we can remove `venv.py` out of `pyodide-build` altogether and maintain it separately. I would be happy to work on this, but will probably do it in a separate repository and shift it here.

Possible use cases:
- allow installing a custom version of `pip` when creating the virtual environment, before it is patched up
- disable `pip` version checks when creating an environment
- embed extra packages into the environment at creation time if placed in a local directory
- and so on

cc: @hoodmane

TODO:

- [x] Examine which args to add
- [x] Add CLI support
- [x] Documentation (most likely self-documented, so we don't need anything else)
- [x] Add some basic tests
	- [x] Figure out how to run the slow integration tests; maybe disable isolation to let them reuse the download xbuildenv

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -50,6 +50,7 @@ jobs:
         with:
           name: coverage-no-integration
           path: .coverage
+          if-no-files-found: error
 
   check-integration-test-trigger:
     name: test-integration-test-trigger
@@ -134,6 +135,7 @@ jobs:
         with:
           name: coverage-integration
           path: .coverage
+          if-no-files-found: error
 
   coverage:
     name: Collect and upload coverage
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,6 +51,7 @@ jobs:
           name: coverage-no-integration
           path: .coverage
           if-no-files-found: error
+          include-hidden-files: true
 
   check-integration-test-trigger:
     name: test-integration-test-trigger
@@ -136,6 +137,7 @@ jobs:
           name: coverage-integration
           path: .coverage
           if-no-files-found: error
+          include-hidden-files: true
 
   coverage:
     name: Collect and upload coverage
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -104,7 +104,7 @@ jobs:
       - name: Install the package
         run: |
           python -m pip install --upgrade pip
-          python -m pip install .
+          python -m pip install -e .
 
       - name: Install xbuildenv
         run: |
@@ -120,20 +120,14 @@ jobs:
         id: get-cores
         run: echo "CORES=$(nproc)" >> $GITHUB_OUTPUT
 
+      - name: Run tests marked with integration
+        run: pytest --junitxml=test-results/junit.xml --cov=pyodide-build pyodide_build -m "integration"
+
       - name: Run the recipe integration tests (${{ matrix.task }})
         env:
           PYODIDE_JOBS: ${{ steps.get-cores.outputs.CORES }}
-        run: |
-          cd integration_tests
-          make ${{ matrix.task }}
-
-      - name: Run tests marked with integration
-        run: |
-          pytest \
-            --junitxml=test-results/junit.xml \
-            --cov=pyodide-build \
-            pyodide_build \
-            -m "integration"
+        working-directory: integration_tests
+        run: make ${{ matrix.task }}
 
       - name: Upload coverage
         uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # v4.6.1
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -38,7 +38,7 @@ jobs:
           python -m pip install --upgrade pip
           pip install -e ".[test]"
 
-          pytest -svra \
+          pytest -v \
             --junitxml=test-results/junit.xml \
             --cov=pyodide-build \
             pyodide_build
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -127,12 +127,13 @@ jobs:
         run: pytest --junitxml=test-results/junit.xml --cov=pyodide-build pyodide_build -m "integration"
 
       - name: Run the recipe integration tests (${{ matrix.task }})
+        if: matrix.task != 'test-integration-marker'
         env:
           PYODIDE_JOBS: ${{ steps.get-cores.outputs.CORES }}
         working-directory: integration_tests
         run: make ${{ matrix.task }}
 
-      - name: Upload coverage
+      - name: Upload coverage for tests marked with integration
         uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # v4.6.1
         if: matrix.task == 'test-integration-marker'
         with:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -85,7 +85,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        task: [test-recipe, test-src]
+        task: [test-recipe, test-src, test-integration-marker]
     if: needs.check-integration-test-trigger.outputs.run-integration-test
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
@@ -123,6 +123,7 @@ jobs:
         run: echo "CORES=$(nproc)" >> $GITHUB_OUTPUT
 
       - name: Run tests marked with integration
+        if: matrix.task == 'test-integration-marker'
         run: pytest --junitxml=test-results/junit.xml --cov=pyodide-build pyodide_build -m "integration"
 
       - name: Run the recipe integration tests (${{ matrix.task }})
@@ -133,8 +134,9 @@ jobs:
 
       - name: Upload coverage
         uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # v4.6.1
+        if: matrix.task == 'test-integration-marker'
         with:
-          name: coverage-integration
+          name: coverage-from-integration
           path: .coverage
           if-no-files-found: error
           include-hidden-files: true
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,11 +13,7 @@ env:
 
 jobs:
   test:
-    strategy:
-      fail-fast: false
-      matrix:
-        os: [ubuntu-latest, macos-latest]
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,6 +16,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          # include tags so that hatch-vcs can infer the version
+          fetch-depth: 0
+          # switch to fetch-tags: true when the following is fixed
+          # see https://github.com/actions/checkout/issues/2041
+          # fetch-tags: true
 
       - name: Setup Python
         uses: actions/setup-python@42375524e23c412d93fb67b49958b491fce71c38 # v5.4.0
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Test
         run: |
           python -m pip install --upgrade pip
-          pip install ".[test]"
+          pip install -e ".[test]"
 
           pytest -v \
             --junitxml=test-results/junit.xml \
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -38,7 +38,7 @@ jobs:
           python -m pip install --upgrade pip
           pip install -e ".[test]"
 
-          pytest -v \
+          pytest -svra \
             --junitxml=test-results/junit.xml \
             --cov=pyodide-build \
             pyodide_build
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,11 @@ env:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -104,7 +104,7 @@ jobs:
       - name: Install the package
         run: |
           python -m pip install --upgrade pip
-          python -m pip install -e .
+          python -m pip install -e ."[test]"
 
       - name: Install xbuildenv
         run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,19 +33,23 @@ jobs:
         with:
           node-version: "20"
 
-      - name: Test
+      - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
           pip install -e ".[test]"
 
-          pytest -v \
+      - name: Run tests
+        run: |
+          pytest \
             --junitxml=test-results/junit.xml \
             --cov=pyodide-build \
             pyodide_build
 
-      - uses: codecov/codecov-action@0565863a31f2c772f9f0395002a31e3f06189574 # v5.4.0
+      - name: Upload coverage
+        uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # v4.6.1
         with:
-          fail_ci_if_error: false
+          name: coverage-no-integration
+          path: .coverage
 
   check-integration-test-trigger:
     name: test-integration-test-trigger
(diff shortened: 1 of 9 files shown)


pyodide_build/out_of_tree/venv.py
```diff
@@ -251,17 +288,27 @@ def install_stdlib(venv_bin: Path) -> None:
...
 
-    if dest.exists():
-        logger.error("ERROR: dest directory '%s' already exists", dest)
-        sys.exit(1)
-
     interp_path = pyodide_dist_dir() / "python"
-    session = session_via_cli(["--no-wheel", "-p", str(interp_path), str(dest)])
+
+    cli_args = ["--python", str(interp_path)]
+
```
agriyakhetarpal, line 299: I changed the default CLI args to also include `wheel` (removed `--no-wheel`). I'm not sure why this was included when this code was written, as I looked at https://github.com/pyodide/pyodide-build/commit/07da853331b6066cb0c77642990a4c6b63337704 and pyodide/pyodide#2976 and couldn't find the reason. Maybe you might be able to recall this, @hoodmane?

Also, I changed `-p` to `--python` as it is more expressive.

pyodide_build/out_of_tree/venv.py
```diff
@@ -251,17 +288,27 @@ def install_stdlib(venv_bin: Path) -> None:
...
+    if virtualenv_args:
+        for arg in virtualenv_args:
+            if arg.startswith("--"):
+                # Check if the argument (or its prefix form) is supported.
+                arg_name = arg.split("=")[0] if "=" in arg else arg
+                if arg_name not in SUPPORTED_VIRTUALENV_OPTIONS:
+                    msg = f"Unsupported virtualenv option: {arg_name}"
+                    logger.warning(msg)
+
+        cli_args.extend(virtualenv_args)
```
agriyakhetarpal, line 309: Should we raise a warning or an error here?

pyodide_build/tests/test_venv.py
```diff
@@ -0,0 +1,225 @@
...
+        assert (temp_venv_path / "bin" / "python").exists()
+        assert (temp_venv_path / "bin" / "pyodide").exists()
+        assert (temp_venv_path / "pip.conf").exists()
+        assert check_function(temp_venv_path)
+
+    except Exception as e:
+        pytest.fail(f"Failed to create virtual environment: {e}")
+
+
+@pytest.mark.skipif(True, reason="Integration test requiring full Pyodide environment")
```
agriyakhetarpal, line 146: These tests are skipped for now, but I'll refactor them later in the day. They are not that slow, but they do take a minute, so I think it makes sense to include them in the integration tests job (one of them also queries PyPI).

pyodide_build/tests/test_venv.py
```diff
@@ -0,0 +1,225 @@
...
+@pytest.mark.parametrize(
+    "packages",
+    [
+        ["six"],  # simple
+        ["numpy"],  # compiled
+        ["six", "numpy"],  # mixed
+    ],
+)
+def test_pip_install(temp_venv_path, packages):
+    """Test that our monkeypatched pip can install packages into the venv"""
```
agriyakhetarpal, line 180: Instead of checking for our `SYSCONFIGDATA` patches and the like, I thought it would be better to see if the end-user functionality is working, as a working `pip` ensures that those patches have worked.

pyodide_build/out_of_tree/venv.py
```diff
@@ -9,6 +9,29 @@
...
+# Our aim will not be to support all of them, and some of them will never be in
+# the list, for example, --no-pip and so on. We provide these on a best-effort
+# basis as they should work and are easy to test.
+SUPPORTED_VIRTUALENV_OPTIONS = [
+    "--clear",
+    "--no-clear",
+    "--no-vcs-ignore",
+    "--symlinks",
+    # "--copies", "--always-copy", FIXME: node fails to invoke Pyodide
+    # "--symlink-app-data", FIXME: node fails to invoke Pyodide
```
agriyakhetarpal, line 22: Noting for completeness: for supporting these two arguments, we'll definitely need to patch the Pyodide Python (which I need to learn more about).

pyodide_build/tests/test_venv.py
agriyakhetarpal: This is resolved with e32a0cca3e8ebda0e7cfc2e6fc3fb12deb533ae6, the tests are a bit faster now.

agriyakhetarpal:
@ryanking13, I'm facing some trouble here where the `dummy_xbuildenv` and `dummy_xbuildenv_url` fixtures that we have in `conftest.py` are conflicting the test fixtures I introduced in the file, which causes the tests to fail when the test suite is run as a whole, but pass when you run the tests separately using `-k test_venv` or similar. I'm curious if you've come across such a situation before? I know about adding [multiple `conftest.py` files](https://docs.pytest.org/en/stable/how-to/fixtures.html#overriding-fixtures-on-various-levels), but I wouldn't like to do that as it will move a few files without any reason.

For context, I'm trying to test the venv creation with multiple options, as I noticed that it hasn't been tested before. It is easy to split `pytest` in CI into two steps, but I'm looking for something more robust so that testing locally remains intact.

ryanking13:
> I'm facing some trouble here where the dummy_xbuildenv and dummy_xbuildenv_url fixtures that we have in conftest.py are conflicting the test fixtures I introduced in the file

Hmm that's weird. Those two fixtures should create xbuildenv in a temporary directory which will be removed after the test function so I am not very sure how they can conflict each other.

Let me see if I can reproduce the error.

ryanking13:
I think I found the root cause. Let me open a PR to fix it.

agriyakhetarpal:
Thank you for the quick fix!

ryanking13:
Looks like the error got different. I guess we need to create a xbuildenv first to create a venv?

ryanking13:
Also, please note that the [dummy_xbuildenv file](https://github.com/pyodide/pyodide-build/tree/main/pyodide_build/tests/_test_xbuildenv) that I made in this repository does not contain the `python` script or other wasm files to run the `pyodide venv` to reduce the size. I think you can update the dummy file to include those files, but it would increase the repository size, so there is a tradeoff.

Some other options would be:

1. Use mock files (with a very small or an empty file) for unittest, and instead use integration tests for testing the "real" `venv` functionality. In integration test, we may use the nightly xbuildenv file from the remote URL.

2. Testing the full functionality of `pyodide venv` in pyodide/pyodide, as the content inside `dist` directory varies per Pyodide version, not by pyodide-build.

3. Adopt `pooch` or other tools as you suggested in https://github.com/pyodide/micropip/issues/165

agriyakhetarpal:
Thanks, Gyeongjae. I'm currently downloading the latest (stable) Pyodide xbuildenv for the package installation tests, which get cached/reused between the tests so that we don't download it for every test. I don't intend to add the real xbuildenv tarball to the repository with this PR – that will be large, as you mention. The CI should pass now, but if you'd like me to label the slow tests as integration tests, I can do that.

agriyakhetarpal:
Okay, almost there – just one failing test (that one also passes on my machine) :D

agriyakhetarpal:
The test fails because it uses the wrong Python interpreter to install packages. It, therefore, installs the macOS wheels for NumPy, and was silently passing in my case. To fix it, we need to call the Pyodide Python interpreter inside the venv (which relies on https://github.com/pyodide/pyodide/pull/5448), or handle the activation/deactivation of the Pyodide venv inside that test to install the packages – I've proceeded with this approach for now in aca03abcb12816e16e53db29d6a6d6cea070ad43.

I see that the rest of the `pyodide venv` functionality is also being tested elsewhere in Pyodide, in this file: https://github.com/pyodide/pyodide/blob/a53c17fd8ef87e79c7d3bfbb1262ca7ff6718ef2/src/tests/test_cmdline_runner.py,  where the tests are more expansive. As @ryanking13 mentioned in https://github.com/pyodide/pyodide/pull/5488, it makes sense to move the relevant tests from that file to `pyodide-build`, so that the functionality is better tested on every commit.

agriyakhetarpal:
Argh, this passes locally and works perfectly fine, but not in CI...

agriyakhetarpal:
So, it passes on macOS indeed, but it fails on Linux.

agriyakhetarpal:
🤔 no luck. I'm unable `python -c "<...>"` doesn't work on Linux. I'll probably drop this part of the test, as it is being tested elsewhere as mentioned in https://github.com/pyodide/pyodide-build/pull/117#issuecomment-2709108527.

pyodide_build/out_of_tree/venv.py
hoodmane: Well there's the question of whether we want `wheel` to run in native Python or in Pyodide's Python. Presumably it should run in the native Python? But that requires extra code to replace the shebang in `bin/wheel`. Also, it's weird magic. It's necessary for pip, but people can run wheel from a normal Python runtime, or if they really want to run it in Pyodide they have the option to `bin/pip install wheel`. So overall, I just don't see the reason for including it.

agriyakhetarpal: Yeah, I think it is always better to run the `wheel` CLI in native Python. I was just interested to know if there was a specific reason not to include it in compatibility terms, so I'll drop it for now. Should we do this for `setuptools` as well using `--no-setuptools`, as I don't see a reason to include it either?

agriyakhetarpal: Never mind, `wheel` and `setuptools` are not being included – I had turned them off in the Typer CLI definitions at some point when working on this PR, so we should be good.

agriyakhetarpal:
> Do we have use cases for all these features? I'm a little reluctant to add the maintenance burden until someone actually wants the missing features. Currently, `pyodide venv` is just good enough for cibuildwheel. In particular, I'm concerned that we might implement the wrong thing, someone might start relying on it, and then we might have to change it later. Perhaps we could mark these options as experimental and subject to breaking changes depending on real world feedback?

Thanks, @hoodmane – your concerns are in the right place, indeed. The primary use case here was to allow debugging #115, and when I thought that we should support `--pip` as a flag, then we could also go all in and support the rest of the options. I've added other possible use cases in the PR description. I think this functionality has a relatively low risk of breaking as we don't do anything wild with `virtualenv`, but I'll mark it as experimental based on your suggestion, as we wouldn't want anyone to depend on these. We can reconsider our stance if we were to publish a `virtualenv-pyodide` plugin.

pyodide_build/out_of_tree/venv.py
```diff
@@ -192,8 +228,9 @@ def create_pip_script(venv_bin):
     ]
 
     for pip in other_pips:
-        pip.unlink()
-        pip.symlink_to(venv_bin / "pip")
+        if pip.exists():
+            pip.unlink()
```
ryanking13, line 232: How about cleaning up all the files in the very first place of the function instead of calling `unlink` in several places?

pyodide_build/tests/test_venv.py
```diff
@@ -0,0 +1,220 @@
...
+    venv_path = base_dir / "test_venv"
+    venv_path.mkdir(exist_ok=True)
+
+    cwd = os.getcwd()
+    os.chdir(str(base_dir))
+
+    xbuildenv_test_name = ".pyodide-xbuildenv-for-testing"
+
+    manager = CrossBuildEnvManager(xbuildenv_test_name)
+    manager.install()
```
ryanking13, line 26: > The CI should pass now, but if you'd like me to label the slow tests as integration tests, I can do that.

Yeah, maybe let's mark tests that rely on this fixture as an integration tests, as I think it is a good to not rely on real xbuildenv in the unittest.

Also calling `install` without the argument assumes that the version will be properly set in the local environment, which is also a bit fragile IMO. So how about passing the URL to the absolute xbuildenv URL instead?

pyodide_build/tests/test_venv.py
```diff
@@ -0,0 +1,220 @@
...
+                # create the directory to avoid cleanup issues
+                temp_venv_path.mkdir(exist_ok=True)
+
+        return MockSession()
+
+    temp_venv_path.mkdir(exist_ok=True)
+
+    # Mock most of the functions called by create_pyodide_venv
+    # as we just need to check the arguments passed to it
+    monkeypatch.setattr("virtualenv.session_via_cli", mock_session_via_cli)
```
ryanking13, line 81: (no need to address in this PR)

It is a bit sad that we need so many mocks there. I guess we can refactor the `venv` code better to provide a more high level interface.

ryanking13 (commented):
Thanks for working on this @agriyakhetarpal! I left some comments.

pyodide_build/tests/test_venv.py
agriyakhetarpal: I agree!

pyodide_build/out_of_tree/venv.py
agriyakhetarpal: Makes sense – addressed in ec3ad527851c6cd66023b1e6201b615f7cb34e39

pyodide_build/tests/test_venv.py
agriyakhetarpal: Fair point – done in commits bf6c872538b1dd0fa96067f2adb924c7bacb9d37 and 6ee83e9c6485f61ede86b051430f5fea346caf66. As we now have two kind of integration tests, where we run the recipes and now some tests that are code-based, I had to split the coverage collection. In another PR, I think we could write a minimal test driver file that would make `pytest` run the `make test-recipe` command, or switch to a task runner designed to work with `pytest`, like [`nox`](https://nox.thea.codes/en/stable/).

agriyakhetarpal:
All actionable feedback addressed for now, @ryanking13. Thanks!

.github/workflows/main.yml
hoodmane, line 1: Are the worklfow changes related to the main changes? Or could we split them into a separate PR?

integration_tests/README.md
hoodmane, line 1: Maybe also split the instruction update into a separate PR since it's not really related.

pyodide_build/cli/venv.py
```diff
@@ -6,12 +6,114 @@
...
+    pip: str = typer.Option(
+        "bundle",
+        "--pip",
+        help="Version of pip to install as seed: embed, bundle, or exact version.",
+    ),
+    setuptools: str | None = typer.Option(
+        None,
+        "--setuptools",
+        help="Version of setuptools to install as seed: embed, bundle, none or exact version",
+    ),
```
hoodmane, line 59: These all seem helpful.

pyodide_build/cli/venv.py
```diff
@@ -6,12 +6,114 @@
...
+    wheel: str | None = typer.Option(
+        None,
+        "--wheel",
+        help="Version of wheel to install as seed: embed, bundle, none or exact version",
+    ),
```
hoodmane, line 64: Let's leave off the `--wheel` option until we decide whether it would run in the native Python or the Pyodide Python.

pyodide_build/cli/venv.py
```diff
@@ -6,12 +6,114 @@
...
+    no_vcs_ignore: bool = typer.Option(
+        False,
+        "--no-vcs-ignore",
+        help="Don't create VCS ignore directive in the destination directory",
+    ),
+    symlinks: bool = typer.Option(
+        True,
+        "--symlinks",
+        help="Try to use symlinks rather than copies, when symlinks are not the default for the platform",
+    ),
```
hoodmane, line 30: For which platforms are symlinks not the default? Windows? We don't support windows at all. I suspect if we don't use a symlink it will be broken because it won't copy some necessary files. Let's leave this one out for now.

hoodmane (approved):
Okay I would like to see the following changes:

- [ ] split workflow changes into separate PR (or explain why we can't)
- [ ] split integration test documentation changes into separate PR (unless new documentation reflects the state of affairs introduced by this PR)
- [ ] Remove --wheel option
- [ ] Remove --symlinks option

Other than that, looks good to me. Thanks @agriyakhetarpal!

pyodide_build/cli/venv.py
agriyakhetarpal: Thanks, dropped it in 3a9a17eb1230f67ac4a14c2a11d5300f62a0ba6f. This commit adds back the `--no-wheel` option, so the previous behaviour has been restored.

pyodide_build/cli/venv.py
agriyakhetarpal: Yes, this is the case for Windows. Though, this option was marked as `True` here and is the default behaviour for Unix-based platforms, so 98c7576086557bf863ca06231d46fe223f94ab2a drops mentions of it in the CLI (so `virtualenv` still uses it, but it isn't overridable now).

agriyakhetarpal:
Thanks to both of you for the reviews!

ryanking13:
Thanks again Agriya! 

It looks like the tests with `integration` marker still run in the unittest not being skipped. Could you please check that part?

agriyakhetarpal:
Oh, I might have missed that part, will take a look!

hoodmane:
This seems to have broken `test_cmdline_runner.py::venv_fail_log[node]`. Specifically, it will override an existing virtual environment now. Before:
```terminal
$ mkdir .venv-pyodide
$ pyodide venv .venv-pyodide
Creating Pyodide virtualenv at .venv-pyodide                                                                                                                                                      
ERROR: dest directory '.venv-pyodide' already exists
```
After:
```terminal
$ mkdir .venv-pyodide
$ pyodide venv .venv-pyodide
Creating Pyodide virtualenv at .venv-pyodide                                                                                                                                                               
... Configuring virtualenv                                                                                                                                                                                 
... Installing standard library                                                                                                                                                                            
Successfully created Pyodide virtual environment!
```

agriyakhetarpal:
Yes, @hoodmane – this was changed when I unlinked the existing symlinks in order to support the `--clear` and `--no-clear` arguments. It brings us closer to `virtualenv`'s behaviour, which also overwrites existing virtual environments if you run the creation twice. My suggestion is to drop that test. Is there a reason we'll want to fail if there's an existing Pyodide venv?

agriyakhetarpal:
Yes, @hoodmane – this was changed when I unlinked the existing symlinks in order to support the `--clear` and `--no-clear` arguments. It brings us closer to `virtualenv`'s behaviour, which also overwrites existing virtual environments if you run the creation twice. My suggestion is to drop/revise that test. Is there a reason we'll want to fail if there's an existing Pyodide venv?

hoodmane:
> Is there a reason we'll want to fail if there's an existing Pyodide venv?

Nope, I just did not realize that the other tools overwrite by default. I will remove the test.

PR #124, opened 2025-03-12
Revert "`skeleton`: predictable URLs for distributions via the PyPI JSON API (#101)"

hoodmane:
The predictable URLs seem to return 404s? @agriyakhetarpal if you could investigate and reland this once we've confirmed that the new URLs work, that would be great.

This reverts commit a16d4681f6cf1e57c2eeeb55011d62dfd56822ad.


agriyakhetarpal:
Thanks, @hoodmane – I tried and I find that it works, but partially. We don't need to completely revert the change as PyPI guarantees by design based on discussions in https://github.com/pypi/warehouse/issues/1944. I identified the bug in my implementation:

For packages with universal (Python 2 and 3) wheels, it gives

`https://files.pythonhosted.org/packages/py2/d/distlib/distlib-0.3.9-py2.py3-none-any.whl`

but it should be

`https://files.pythonhosted.org/packages/py2.py3/d/distlib/distlib-0.3.9-py2.py3-none-any.whl`

hoodmane:
Thanks for the quick fix @agriyakhetarpal. I am not sure, I think you're more likely to notice any problems than me. Anyways I will close this if you already think you know how to fix it.

PR #129, opened 2025-03-13
Virtual environment: handle pip updates

### LINKED ISSUE #128: Add `pyodide venv --repair [DEST]` option
As I noticed in #115, re-installing `pip` will break completely and thus break the `venv`, as if `pip` isn't working, it is not possible to install or uninstall any packages. I wonder if we could support a `--apply-pip-monkeypatch` or `--repair` option that just applies the patch to fix a broken venv.

hoodmane:
Instead of putting our payload into pip and making pip3 and pip3.12 symlinks to pip, put our payload into pip_patched and symlink all pip variants to that. Then we can check for a problematic update by testing whether readlink() of the current executing file still points to pip_patched. If it doesn't, put it back.

Resolves #128

### DIFF
--- a/pyodide_build/out_of_tree/venv.py
+++ b/pyodide_build/out_of_tree/venv.py
@@ -174,6 +174,45 @@ def get_pip_monkeypatch(venv_bin: Path) -> str:
         sysconfig._init_config_vars()
         del os.environ["_PYTHON_SYSCONFIGDATA_NAME"]
         """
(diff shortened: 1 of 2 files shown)


agriyakhetarpal (approved):
Awesome! Thanks for the quick patch, @hoodmane (no pun intended :D)

hoodmane:
Thanks for the review @agriyakhetarpal

PR #131, opened 2025-03-13
Add testing for macOS

### LINKED ISSUE #127: Add testing on macOS
Currently, we don't have testing on macOS, and I see a few tests failing on my laptop locally. I think it would be a good idea to run the test suite on both Linux and macOS, as `cibuildwheel` uses a bunch of `pyodide`'s functionality here and has advertised testing on macOS (as "should work but is not tested", if I remember correctly).

agriyakhetarpal:
## Description

This pull request closes #127 by adding macOS as a tested platform., which should help our clients such as `cibuildwheel` and also aid local testing for contributors on macOS devices. We could skip the integration tests for macOS, but I think we would be fine with keeping them as we would still be under the limit for the number of concurrent jobs on a free GHA plan (8).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,10 @@ env:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
@@ -81,12 +84,13 @@ jobs:
           fi
 
   integration-test:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     needs: [check-integration-test-trigger]
     strategy:
       fail-fast: false
       matrix:
         task: [test-recipe, test-src, test-integration-marker]
+        os: [ubuntu-latest, macos-latest]
     if: needs.check-integration-test-trigger.outputs.run-integration-test
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
--- a/pyodide_build/tests/recipe/test_bash_runner.py
+++ b/pyodide_build/tests/recipe/test_bash_runner.py
@@ -1,3 +1,4 @@
+import os
 import subprocess
 from pathlib import Path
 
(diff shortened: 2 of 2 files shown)


pyodide_build/tests/recipe/test_bash_runner.py
```diff
@@ -53,11 +53,17 @@ def test_subprocess_with_shared_env_logging(capfd, tmp_path):
...
+        # Clean and join all outputs to handle potential line breaks,
+        # and check for path components, instead of the exact path.
+        assert "Running test script in" in cap.out
+
+        # Normalize whitespace for path comparison
+        normalized_output = "".join(cap.out.split())
+        normalized_path = "".join(str(dir).split())
+        assert normalized_path in normalized_output
+
+        assert "1000" in cap.out
```
agriyakhetarpal, line 66: Unfortunately, macOS runs with `zsh` and not `bash`, which forces us to resort to things like these. Changing to `bash` does not help, either.

A similar observation was noted in #105.

pyodide_build/tests/recipe/test_bash_runner.py
```diff
@@ -53,11 +53,17 @@ def test_subprocess_with_shared_env_logging(capfd, tmp_path):
...
-            str(dir),
-            "1000",
-        ]
+
+        # Clean and join all outputs to handle potential line breaks,
+        # and check for path components, instead of the exact path.
+        assert "Running test script in" in cap.out
+
+        # Normalize whitespace for path comparison
+        normalized_output = "".join(cap.out.split())
```
hoodmane, line 62: The old code is significantly easier to read. What goes wrong with `str(dir)` on mac? Could you only normalize the output and not what we are comparing it to? Like:
```py
normalized_dir = str(dir)
assert normalize_output(cap.out) == [ "Running test script in", normalized_dir, "1000" ]
```
and them make `normalize_output` do whatever you need to clean it up?

pyodide_build/tests/recipe/test_bash_runner.py
agriyakhetarpal: Ah, I haven't explored why it doesn't help; it's just that I encounter the same failure.

.github/workflows/main.yml
```diff
@@ -155,4 +159,4 @@ jobs:
       - uses: codecov/codecov-action@0565863a31f2c772f9f0395002a31e3f06189574 # v5.4.0
         with:
           fail_ci_if_error: false
-          files: coverage-no-integration/.coverage,coverage-integration/.coverage
+          files: coverage-no-integration-ubuntu-latest/.coverage,coverage-integration-ubuntu-latest/.coverage,coverage-no-integration-macos-latest/.coverage,coverage-integration-macos-latest/.coverage
```
ryanking13, line 162: It is unfortunate that we cannot use the parameters in the matrix here. But I don't think it's a big deal if we need to fix two places. Could you please add a comment that when fixing the matrix, one should fix the names here as well?

ryanking13 (approved):
Thanks!

.github/workflows/main.yml
agriyakhetarpal: Makes sense to me. I think this could be simplified by running `coverage combine` over the various coverage files. Let me see if that works, as the current approach is indeed prone to mishaps, like you suggest.

agriyakhetarpal:
Thanks for the reviews, @hoodmane and @ryanking13!

PR #132, opened 2025-03-13
Add initial support for uv

### LINKED ISSUE #58: `pyodide xbuildenv install` won't work if installed in `uv`-managed virtual environments

ryanking13:
Adds initial support for uv: make `uv run pyodide ...` work. Resolve #58

- [x] changelog


pyproject.toml
```diff
@@ -63,6 +63,9 @@ test = [
     "pytest-cov",
     "types-requests",
 ]
+uv = [
+  "build[uv]~=1.2.0",
+]
```
agriyakhetarpal, line 68: Side note; I remember that `pypa/build` 1.2.0 broke things for us last year. Would it be worth it for us to vendor `pypa/build`, so that `pypabuild.py` can use it instead?

agriyakhetarpal (approved):
Thanks, @ryanking13! The failing test on macOS looks like an easy fix. Also, marking #58 as related (could you please check if it fully resolves the issue?)

pyodide_build/xbuildenv.py
```diff
@@ -265,15 +265,28 @@ def _install_cross_build_packages(
...
+
+        install_prefix = (
             [
+                uv_helper.find_uv_bin(),
+                "pip",
+                "install",
+            ]
+            if uv_helper.should_use_uv()
+            else [
                 "pip",
```
agriyakhetarpal, line 277:
```suggestion
            else [
                sys.executable,
                "-m"
                "pip",
```

I missed this in my previous review above, from #58

pyodide_build/recipe/graph_builder.py
```diff
@@ -130,8 +130,12 @@ def needs_rebuild(self, build_dir: Path) -> bool:
         return res
 
     def build(self, build_args: BuildArgs, build_dir: Path) -> None:
+        run_prefix = (
+            [uv_helper.find_uv_bin(), "run"] if uv_helper.should_use_uv() else []
+        )
```
agriyakhetarpal, line 135: We should be careful about this, based on https://github.com/pyodide/pyodide-build/pull/132/files#r1996940827.

ryanking13: Hmm, I see. Probably, we should entirely switch from subprocess to multiprocessing to avoid this kind of problem.

pyproject.toml
ryanking13: > Side note; I remember that pypa/build 1.2.0 broke things for us last year.

I don't remember this. Could you please elaborate? Note that `build~=1.2.0` is also set in the dependencies.

> Would it be worth it for us to vendor pypa/build, so that pypabuild.py can use it instead?

No, I don't think vendoring is a good idea for `pypa/build` case. If we want to pin into a specific version, I would choose to pin the version in pyproject.toml instead.

ryanking13:
Okay, for now I will go ahead and merge. Let's keep discuss how we can improve uv support.

pyproject.toml
agriyakhetarpal: Ah, sounds good to me, thanks! It was this issue: https://github.com/numpy/numpy/issues/26164

PR #133, opened 2025-03-13
Set wheel tag of cpython_module to cp312-cp312-pyodide_2025_0_wasm32

hoodmane:
We were tagging them py3-none-any which is not correct.

### DIFF
--- a/pyodide_build/common.py
+++ b/pyodide_build/common.py
@@ -334,7 +334,18 @@ def modify_wheel(wheel: Path) -> Iterator[Path]:
         pack_wheel(wheel_dir, wheel.parent)
 
 
-def retag_wheel(wheel_path: Path, platform: str) -> Path:
+def retag_wheel(
+    wheel_path: Path,
+    platform: str,
+    *,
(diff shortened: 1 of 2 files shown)


hoodmane:
I'll check if it's simple to test.

hoodmane:
I think it shouldn't be too hard to add to `_test_recipes` a cpython_module recipe.

ryanking13:
> I think it shouldn't be too hard to add to _test_recipes a cpython_module recipe.

Yeah, I think you can add one there

hoodmane:
Okay I added a test. Thanks for all your hard work adding test coverage to pyodide-build @ryanking13. The test suite is quite nice.

agriyakhetarpal (approved):

hoodmane:
@agriyakhetarpal any idea why the added test fails on the macos runner?

hoodmane:
Well possibly it fails on both, can we turn off fail-fast?

hoodmane:
And of course the problem is I forgot `git add`.

PR #134, opened 2025-03-13
Make needs_rebuild aware of wheel tags

hoodmane:
This switches to using the wheel itself instead of the ".packaged" marker to decide if a wheel needs to be rebuilt. Static and dynamic libraries still use the ".packaged" marker.

This solves the problem that whenever I switch branches between a Python update branch and any other branch, absolutely everything has to be rebuilt. Instead, this saves the wheels from each branch next to each other.

This should also resolve the annoying `Found multiple matching wheels` errors when switching abis or Python versions.

### DIFF
--- a/pyodide_build/build_env.py
+++ b/pyodide_build/build_env.py
@@ -6,7 +6,7 @@ import os
 import re
 import subprocess
 import sys
-from collections.abc import Iterator
+from collections.abc import Iterator, Sequence
 from contextlib import nullcontext, redirect_stdout
 from io import StringIO
 from pathlib import Path
@@ -219,7 +219,7 @@ def wheel_platform() -> str:
     return f"pyodide_{abi_version}_wasm32"
 
 
-def pyodide_tags() -> Iterator[Tag]:
+def pyodide_tags_() -> Iterator[Tag]:
     """
     Returns the sequence of tag triples for the Pyodide interpreter.
 
@@ -233,6 +233,11 @@ def pyodide_tags() -> Iterator[Tag]:
     yield from compatible_tags(platforms=PLATFORMS, python_version=python_version)
     # Following line can be removed once packaging 22.0 is released and we update to it.
     yield Tag(interpreter=f"cp{PYMAJOR}{PYMINOR}", abi="none", platform="any")
+
+
+@functools.cache
+def pyodide_tags() -> Sequence[Tag]:
+    return list(pyodide_tags_())
 
 
 def replace_so_abi_tags(wheel_dir: Path) -> None:
--- a/pyodide_build/common.py
+++ b/pyodide_build/common.py
@@ -14,7 +14,7 @@ import tomllib
 import warnings
 import zipfile
 from collections import deque
-from collections.abc import Generator, Iterable, Iterator, Mapping
+from collections.abc import Generator, Iterable, Iterator, Mapping, Sequence
 from contextlib import contextmanager
 from pathlib import Path
 from tempfile import NamedTemporaryFile, TemporaryDirectory
@@ -35,11 +35,13 @@ def xbuildenv_dirname() -> str:
     return f".pyodide-xbuildenv-{__version__}"
 
 
-def find_matching_wheels(
-    wheel_paths: Iterable[Path], supported_tags: Iterator[Tag]
+def _find_matching_wheels(
+    wheel_paths: Iterable[Path], supported_tags: Sequence[Tag], version: str = None
 ) -> Iterator[Path]:
     """
     Returns the sequence wheels whose tags match the Pyodide interpreter.
+
+    We don't bother ordering them carefully because we are only hoping to find one.
 
     Parameters
     ----------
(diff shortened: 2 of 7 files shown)


pyodide_build/recipe/builder.py
```diff
@@ -523,8 +529,7 @@ def _build_package(self, bash_runner: BashRunnerWithSharedEnvironment) -> None:
             self._compile(bash_runner)
 
         self._package_wheel(bash_runner)
-        shutil.rmtree(self.dist_dir, ignore_errors=True)
-        shutil.copytree(self.src_dist_dir, self.dist_dir)
+        shutil.copytree(self.src_dist_dir, self.dist_dir, dirs_exist_ok=True)
```
hoodmane, line 532: Leave the old wheels instead of removing them. They can be reused on a branches where they still have the appropriate package version, python version, and abi version.

pyodide_build/recipe/builder.py
```diff
@@ -798,14 +808,22 @@ def needs_rebuild(
...
+    if is_wheel:
+        previous_wheel = find_matching_wheel(
+            dist_dir.glob("*.whl"), pyodide_tags(), version=version
         )
-        return True
-
-    package_time = packaged_token.stat().st_mtime
+        if not previous_wheel:
+            return True
+        package_time = previous_wheel.stat().st_mtime
```
hoodmane, line 818: For wheels use the wheel itself to determine the modification time for deciding if a rebuild is needed.

pyodide_build/build_env.py
```diff
@@ -235,6 +235,11 @@ def pyodide_tags() -> Iterator[Tag]:
     yield Tag(interpreter=f"cp{PYMAJOR}{PYMINOR}", abi="none", platform="any")
 
 
+@functools.cache
+def pyodide_tags() -> Sequence[Tag]:
+    return list(pyodide_tags_())
```
ryanking13, line 240: Why not just cache the original `pyodide_tags` directly?

pyodide_build/common.py
```diff
@@ -35,12 +35,14 @@ def xbuildenv_dirname() -> str:
     return f".pyodide-xbuildenv-{__version__}"
 
 
-def find_matching_wheels(
-    wheel_paths: Iterable[Path], supported_tags: Iterator[Tag]
+def _find_matching_wheels(
+    wheel_paths: Iterable[Path], supported_tags: Sequence[Tag], version: str = None
```
ryanking13, line 39:
```suggestion
    wheel_paths: Iterable[Path], supported_tags: Sequence[Tag], version: str | None = None
```

pyodide_build/recipe/builder.py
```diff
@@ -176,7 +177,8 @@ def build(self) -> None:
         try:
             self._build()
 
-            (self.build_dir / ".packaged").touch()
+            if not self.is_wheel:
+                (self.build_dir / ".packaged").touch()
```
ryanking13, line 181: I think it could be improved by implementing it in each child class. You don't need to implement it in this PR, and since I was working on subclassing, I can do it all at once in another PR.

pyodide_build/recipe/builder.py
```diff
@@ -541,12 +546,13 @@ def _package_wheel(
...
         """
-        wheel, *rest = find_matching_wheels(
-            self.src_dist_dir.glob("*.whl"), pyodide_tags()
+        wheel = find_matching_wheel(
+            self.src_dist_dir.glob("*.whl"), pyodide_tags(), version=self.version
         )
-        if rest:
-            raise Exception(
-                f"Unexpected number of wheels {len(rest) + 1} when building {self.name}"
+        if not wheel:
```
ryanking13, line 552: 👍  Looks much better.

pyodide_build/tests/test_cli.py
```diff
@@ -23,6 +23,16 @@
 RECIPE_DIR = Path(__file__).parent / "recipe" / "_test_recipes"
 
 
+def assert_runner_succeeded(result):
+    __tracebackhide__ = True
+    print(result.stdout)
```
ryanking13, line 28:
```suggestion
```

ryanking13 (approved):
Thanks!

pyodide_build/tests/test_cli.py
hoodmane: I think this is helpful? In my opinion, either the test is passing and stdout gets hidden or its failing and stdout is useful.

pyodide_build/tests/recipe/test_builder.py
```diff
@@ -248,11 +248,23 @@ def _check_patches_extra(self) -> Self:
...
+    version = "12"
+    if is_wheel:
+        dist_dir = pkg_root / "dist"
+        dist_dir.mkdir()
+        # Build of current version with wrong abi
+        (dist_dir / "regex-12-cp311-cp311-pyodide_2024_0_wasm32.whl").touch()
+        # Build of old version with current abi
+        (dist_dir / "regex-11-cp312-cp312-pyodide_2024_0_wasm32.whl").touch()
+        # the version we're trying to build
+        packaged = dist_dir / "regex-12-cp312-cp312-pyodide_2024_0_wasm32.whl"
```
agriyakhetarpal, line 265:
```suggestion
        # Build of current version with wrong abi
        (dist_dir / "regex-12-cp311-cp311-pyodide_2024_0_wasm32.whl").touch()
        # Build of old version with current abi
        (dist_dir / "regex-11-cp312-cp312-pyodide_2025_0_wasm32.whl").touch()
        # the version we're trying to build
        packaged = dist_dir / "regex-12-cp312-cp312-pyodide_2024_0_wasm32.whl"
```

I think this might have been a typo? Please feel free to discard otherwise.

pyodide_build/tests/test_cli.py
agriyakhetarpal: I, too, think that this would be helpful – though it would be hidden anyway unless we use the `-s` option with `pytest`; [we currently use `-vra`](https://github.com/pyodide/pyodide-build/blob/147c1557adaccb5ab2356b7ed4c334edb04c7596/pyproject.toml#L76).

agriyakhetarpal (approved):
Thanks, @hoodmane!

pyodide_build/tests/test_cli.py
hoodmane: Yeah it's preferable for it to be hidden by default.

pyodide_build/tests/recipe/test_builder.py
hoodmane: I think this test shims the abi tag so the "current" abi is actually whatever is specified here. In any case, the test passes as written.

PR #135, opened 2025-03-15
Apply PIP_CONSTRAINT while installing dependencies

hoodmane:
We apply the PIP_CONSTRAINT by setting an environment variable. But it isn't around when we actually install dependencies so it is ignored. This fixes it.

### DIFF
--- a/pyodide_build/build_env.py
+++ b/pyodide_build/build_env.py
@@ -318,7 +318,8 @@ def _create_constraints_file() -> str:
 
     if len(constraints.split(maxsplit=1)) > 1:
         raise ValueError(
-            "PIP_CONSTRAINT contains spaces so pip will misinterpret it. Make sure you clone Pyodide on a path with no spaces."
+            "PIP_CONSTRAINT contains spaces so pip will misinterpret it. Make sure the path to pyodide has no spaces.\n"
+            "See https://github.com/pypa/pip/issues/13283"
         )
 
     constraints_file = Path(constraints)
--- a/pyodide_build/build_env.py
+++ b/pyodide_build/build_env.py
@@ -316,6 +316,11 @@ def _create_constraints_file() -> str:
     if not constraints:
         return ""
 
+    if len(constraints.split(maxsplit=1)) > 1:
+        raise ValueError(
+            "PIP_CONSTRAINT contains spaces so pip will misinterpret it. Make sure you clone Pyodide on a path with no spaces."
+        )
+
     constraints_file = Path(constraints)
     if not constraints_file.is_file():
         constraints_file.parent.mkdir(parents=True, exist_ok=True)
--- a/pyodide_build/build_env.py
+++ b/pyodide_build/build_env.py
@@ -305,3 +305,19 @@ def local_versions() -> dict[str, str]:
         "pyodide-build": __version__,
         # "emscripten": "TODO"
     }
+
+
+def _create_constraints_file() -> str:
+    try:
+        constraints = get_build_flag("PIP_CONSTRAINT")
+    except ValueError:
+        return ""
+
+    if not constraints:
+        return ""
+
+    constraints_file = Path(constraints)
+    if not constraints_file.is_file():
+        constraints_file.parent.mkdir(parents=True, exist_ok=True)
+        constraints_file.write_text("")
+    return constraints
--- a/pyodide_build/pypabuild.py
+++ b/pyodide_build/pypabuild.py
@@ -160,7 +160,7 @@ def _build_in_isolated_env(
         # first install the build dependencies
         symlink_unisolated_packages(env)
         install_reqs(build_env, env, builder.build_system_requires)
-        build_reqs = None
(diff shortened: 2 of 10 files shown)


pyodide_build/pypabuild.py
```diff
@@ -142,7 +148,7 @@ def _build_in_isolated_env(
     # For debugging: The following line disables removal of the isolated venv.
     # It will be left in the /tmp folder and can be inspected or entered as
     # needed.
-    # _DefaultIsolatedEnv.__exit__ = lambda *args: None
+    # _DefaultIsolatedEnv.__exit__ = lambda self, *args: print("Skipping removing isolated env in", self.path)
```
hoodmane, line 151: We should add a cli flag that enables this.

pyodide_build/build_env.py
```diff
@@ -305,3 +305,19 @@ def local_versions() -> dict[str, str]:
...
+        return ""
+
+    if not constraints:
+        return ""
+
+    constraints_file = Path(constraints)
+    if not constraints_file.is_file():
+        constraints_file.parent.mkdir(parents=True, exist_ok=True)
+        constraints_file.write_text("")
+    return constraints
```
agriyakhetarpal, line 323:
```suggestion
def _create_constraints_file() -> str:
    try:
        constraints = get_build_flag("PIP_CONSTRAINT")
    except ValueError:
        return ""

    if not constraints:
        return ""

    constraints_file = Path(constraints).resolve()
    if not constraints_file.is_file():
        constraints_file.parent.mkdir(parents=True, exist_ok=True)
        constraints_file.write_text("")
    return constraints
```

`resolve()`ing the `Path` should also make it track symlinks, and I feel it is generally better, unless there is a special case.

Also, a question: [the `pip` documentation](https://pip.pypa.io/en/latest/cli/pip_install/#cmdoption-c) says that it can be supplied multiple times. Is that also possible through the environment variable?

agriyakhetarpal (commented):
Thanks, @hoodmane, please let me know if when it is fully ready for review!

hoodmane:
It's ready for review.

pyodide_build/pypabuild.py
```diff
@@ -153,26 +159,29 @@ def _build_in_isolated_env(
 
         # first install the build dependencies
         symlink_unisolated_packages(env)
-        install_reqs(env, builder.build_system_requires)
-        installed_requires_for_build = False
+        install_reqs(build_env, env, builder.build_system_requires)
+        build_reqs = None
```
ryanking13, line 163: Could you please add a type annotation here?

pyodide_build/pypabuild.py
```diff
@@ -153,26 +159,29 @@ def _build_in_isolated_env(
...
-            install_reqs(env, build_reqs)
-            installed_requires_for_build = True
 
-        with common.replace_env(build_env):
-            if not installed_requires_for_build:
+        if not build_reqs:
+            # get_requires_for_build in native env failed. Maybe trying to
+            # execute get_requires_for_build in the cross build environment will
+            # work?
+            # TODO: Is this case ever used? Add test coverage or remove.
```
ryanking13, line 175: I saw that this case was used in pygame-ce. In native env, the setup.py picks up native SDL2 config, then fails. In the cross env, it correctly picks up Emscripten SDL2 config.

[pygame-ce.log](https://github.com/user-attachments/files/19269447/pygame-ce.log)

ryanking13 (approved):
Thanks!

pyodide_build/build_env.py
hoodmane: It seems that it is possible, using spaces to delimit multiple entries:
https://github.com/pypa/pip/blob/main/src/pip/_internal/cli/parser.py?plain=1#L248-L250

As a corollary to this, it is NOT possible to specify via the environment variable any path with spaces in it. But hopefully nobody will clone Pyodide into a folder with spaces in the name. Maybe I'll put a check for it.

hoodmane: Opened an issue in pip:
https://github.com/pypa/pip/issues/13283

pyodide_build/pypabuild.py
hoodmane: Updated the comment.

hoodmane: I guess a second question is whether the first attempt is ever necessary. It would be nice to only have one of these. But we can investigate later.

ryanking13: Well, packages that are unaware of Emscripten would have something like:

```python
match sys.platform:
  case 'linux': do_somthing_linux()
  case 'macos': do_somthing_macos()
  case 'windows': dos_something_windows()
  default: raise RuntimeError("now supported platform")
```

and in this case, we would like to choose linux for a best effort.

hoodmane: In that case, maybe we could move towards selecting which of these behaviors is desired in the `meta.yaml` file. It would be nice to be more explicit about this. But perhaps no action is needed.

PR #137, opened 2025-03-16
Add earlier check for wheel version mismatch

hoodmane:
And a more descriptive error message

### DIFF
--- a/pyodide_build/recipe/builder.py
+++ b/pyodide_build/recipe/builder.py
@@ -16,6 +16,7 @@ from pathlib import Path
 from typing import Any, cast
 
 import requests
+from packaging.utils import parse_wheel_filename
 
 from pyodide_build import common, pypabuild
 from pyodide_build.build_env import (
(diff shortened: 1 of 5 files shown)


pyodide_build/recipe/builder.py
```diff
@@ -82,6 +83,14 @@ def _extract_tarballname(url: str, headers: dict) -> str:
     return tarballname
 
 
+def check_versions_match(wheel_name: str, version: str):
+    wheel_version = str(parse_wheel_filename(wheel_name)[1])
+    if wheel_version != version:
+        raise ValueError(
+            f"Version mismatch: version in meta.yaml is '{version}' but version from wheel name is '{wheel_version}'"
+        )
```
ryanking13, line 91: how about including the `wheel_name` in the error message as well?

pyodide_build/recipe/builder.py
```diff
@@ -332,6 +341,7 @@ def _download_and_extract(self) -> None:
 
         # already built
         if tarballpath.suffix == ".whl":
+            check_versions_match(tarballpath.name, self.version)
```
ryanking13, line 344: If we don't want to download the wheel file, I think it's possible to check before it is downloaded by looking at the URL.

ryanking13 (approved):
Thanks!

PR #141, opened 2025-03-18
Set CMAKE_EXECUTABLE env variable to make scikit-build use cmake wrapper

ryanking13:
Previously, we were skipping `cmake` build dependency becuase installing it would shadow the cmake pywasmcross wrapper. However, scikit-build introduced `CMAKE_EXECUTABLE` env variable to locate the cmake binary, so I think we can utilize that.

- [x] test in pyodide/pyodide before merge
- [x] changelog


ryanking13:
It seems like we have an issue with the macros + scikit-build-core combination. It isn't related to this PR but I guess I'll need to fix that first.

hoodmane (approved):
Looks good to me. One question: Do we have a way to add caps to uncapped cmake versions requested by packages?

ryanking13:
> uncapped cmake versions requested by packages?

As you've fixed PIP_CONSTRAINTS issue #135, I think we can put `cmake < 4.0` in the `requirement.constraint` field or use constraints.txt file to apply it globally.

integration_tests/recipes/boost-histogram/meta.yaml
```diff
@@ -0,0 +1,19 @@
+package:
+  name: boost-histogram
+  version: 1.5.0
+  top-level:
+    - boost_histogram
+source:
+  url: https://files.pythonhosted.org/packages/8d/51/96712ed4e1d64618d3e76aac2d1f96e2f4856120012f0dad51adf42c78e4/boost_histogram-1.5.0.tar.gz
```
agriyakhetarpal, line 7:
```suggestion
  url: https://files.pythonhosted.org/packages/source/b/boost_histogram/boost_histogram-1.5.0.tar.gz
```

(nit)

agriyakhetarpal (approved):
Thanks, @ryanking13, looks good to me!

I had a few questions:
- should we also test a package that has a classic `setuptools` + CMake-based build, outside of `scikit-build-core`?
- how does `cibuildwheel` not have the problem on macOS – is it intercepting and renaming the platform tag for the wheel to the right one? If yes, should we do so too?

ryanking13:
> how does cibuildwheel not have the problem on macOS – is it intercepting and renaming the platform tag for the wheel to the right one? If yes, should we do so too?

Sounds good. If there is a good candidate, could you please add one? opencv-python can be a good candidate that does not use scikit-build-core (it use the legacy scikit-build) but it is too heavy to build.

> how does cibuildwheel not have the problem on macOS – is it intercepting and renaming the platform tag for the wheel to the right one? If yes, should we do so too?

I have no idea. Are we sure that cibuildwheel + macOS + scikit-build-core works? Is there a good example for that case?

ryanking13:
> Yes, cibuildwheel + macOS + scikit-build-core works – it worked the last time I tried it a couple of months ago. Any package that uses scikit-build-core (that we have a recipe for) would work well.

The problem is that on macOS + scikit-build-core combination, the wheel file is outputted with native wheel name. In  `pyodide build-recipes` we expect the wheel filename with the emscripten tag, hence the problem. Meanwhile, `cibuildwheel` uses `pyodide build` and does not check the output filename. So I think that is why cibuildwheel works.

One question is, in that case, does `cibuildwheel` rename the built wheel file? Or does it leave the file as-is? If the latter is the case, the output wheel of cibuildwheel + macOS + scikit-build-core combination would still have the native wheel name.

agriyakhetarpal:
Having looked at `cibuildwheel` in https://github.com/pypa/cibuildwheel/blob/4ef7b3770f3ebfdd8025f269b16c3dd5d4080481/cibuildwheel/platforms/pyodide.py, it doesn't rename the file. I might have mistaken myself with a CI run and something that I ran locally which did not use `scikit-build-core`. Sorry for the confusion!

I think we could check for the magic number for all the compiled `.so`s for the wheel, and if it matches with `\0asm` for all of them, rename the wheel ourselves – declaring that the wheel is indeed a WASM wheel? Or, would this be a hack, and out of scope for us?

ryanking13:
> I think we could check for the magic number for all the compiled .sos for the wheel, and if it matches with \0asm for all of them, rename the wheel ourselves – declaring that the wheel is indeed a WASM wheel? Or, would this be a hack, and out of scope for us?

Maybe, yes. But that is the opposite direction of our recent approach (https://github.com/pyodide/pyodide-build/pull/134), so we will have to discuss about it.

PR #144, opened 2025-03-20
Make pyodide venv work from a uv-managed python virtual environment

### LINKED ISSUE #143: pip fails with `No module named 'encodings'` in pyodide venv created from uv python
This is coming from https://discuss.python.org/t/pep-776-emscripten-support/84996/20

I see an error when following the instructions from the docs to create venv and use pip:
https://pyodide.org/en/stable/development/building-and-testing-packages.html

It is possible that this is due to using a uv installed Python.

Specifically I run these commands that seem to complete successfully until the last:
```
uv python install 3.12
uv venv -p 3.12 .venv
source .venv/bin/activate
uv pip install pip
pip install pyodide-build
pyodide venv .venv_pyodide
source .venv_pyodide/bin/activate
pip # error
```
The error I see is (I have redacted two paths):
```
$ pip
Could not find platform independent libraries <prefix>
Could not find platform dependent libraries <exec_prefix>
Python path configuration:
  PYTHONHOME = (not set)
  PYTHONPATH = (not set)
  program name = '.../.venv_pyodide/bin/python3.12-host'
  isolated = 0
  environment = 1
  user site = 0
  safe_path = 0
  import site = 1
  is in build tree = 0
  stdlib dir = '/install/lib/python3.12'
  sys._base_executable = '/home/oscar/.local/share/uv/python/cpython-3.12.9-linux-x86_64-gnu/bin/python3.12'
  sys.base_prefix = '/install'
  sys.base_exec_prefix = '/install'
  sys.platlibdir = 'lib'
  sys.executable = '.../.venv_pyodide/bin/python3.12-host'
  sys.prefix = '/install'
  sys.exec_prefix = '/install'
  sys.path = [
    '/install/lib/python312.zip',
    '/install/lib/python3.12',
    '/install/lib/python3.12/lib-dynload',
  ]
Fatal Python error: init_fs_encoding: failed to get the Python codec of the filesystem encoding
Python runtime state: core initialized
ModuleNotFoundError: No module named 'encodings'

Current thread 0x00007b7352aeb740 (most recent call first):
  <no Python frame>
```
I have not tested this yet with a non-uv Python.

hoodmane:
Symlinking python-host to the uv python causes trouble, so instead we use exec.

Now the following works:
```sh
uv python install 3.12
uv venv -p 3.12 .venv
source .venv/bin/activate
uv pip install pip pyodide-build
pyodide venv .venv_pyodide
.venv_pyodide/bin/pip --help
```

Resolves #143. 

It is still necessary to do `uv pip install pip` first, that is a different bug.

- [ ] Add a test

### DIFF
--- a/pyodide_build/out_of_tree/venv.py
+++ b/pyodide_build/out_of_tree/venv.py
@@ -234,9 +234,19 @@ def create_pip_script(venv_bin):
         pip.unlink(missing_ok=True)
         pip.symlink_to(pip_path)
 
-    host_python_path.symlink_to(sys.executable)
+    # We used to use a symlink here, but getpath.py failed when used with uv
+    # python. So now we do this instead.
+    host_python_path.write_text(
+        dedent(
+            f"""\
+            #!/bin/sh
+            exec {sys.executable} $@
+            """
+        )
+    )
+    host_python_path.chmod(0o777)
     # in case someone needs a Python-version-agnostic way to refer to python-host
-    (venv_bin / "python-host").symlink_to(sys.executable)
+    (venv_bin / "python-host").symlink_to(host_python_path)
 
     pip_path.write_text(
         # Other than the shebang and the monkey patch, this is exactly what


agriyakhetarpal (approved):
Nice, thanks, @hoodmane! There's no way to avoid `uv pip install pip`, or can there be?

Could you please add a CHANGELOG entry (which mentions that `pip` still needs to be installed)?

hoodmane:
> There's no way to avoid uv pip install pip, or can there be?

I haven't looked into it, we could make a separate issue for that.

agriyakhetarpal:
Hi Hood, should we include this PR in my next release? I just wanted to know about it, as you didn't merge it following my approval, so maybe you're waiting for something here. I think it should be good to go with a CHANGELOG and the note about the caveat (in the CHANGELOG, but perhaps also in code when going through the `should_use_uv()` code path).

I also have to note that `uv`-based virtual environments won't add `pip` to PATH even if we do `uv pip install pip` to prevent unexpected behaviour, so maybe that's something we should handle?

hoodmane:
I was waiting to add a test but we could merge without one.

> I also have to note that uv-based virtual environments won't add pip to PATH even if we do uv pip install pip to prevent unexpected behaviour, so maybe that's something we should handle?

It worked fine for me with this change and `uv pip install pip` so I'm not sure that there's a problem.

ryanking13:
> Hi Hood, should we include this PR in my next release? I just wanted to know about it,

I think you don't need to wait for other PRs to release. Actually, putting smaller changes in a single release is better for us to prevent any kind of regression.

Issue #114, opened 2025-03-03
Support `--xbuildenv-path` flag and `PYODIDE_XBUILDENV_PATH` environment variable with `pyodide build`


agriyakhetarpal:
(Related to #72), and as stated in the issue title. I think we should allow this, as `pyodide build` at the moment downloads multiple cross-build environments (in whichever folder the command is run from, rather than using the existing one). If `pyodide build` can know where to look, we can specify that and not use excess bandwidth.

My question is, should we call it `--xbuildenv-path`/`PYODIDE_XBUILDENV_PATH` or `--xbuildenv-dir`/`PYODIDE_XBUILDENV_DIR`, i.e., `path` vs `dir`?

ryanking13:
> My question is, should we call it --xbuildenv-path/PYODIDE_XBUILDENV_PATH or --xbuildenv-dir/PYODIDE_XBUILDENV_DIR, i.e., path vs dir?

Not a strong preference, but I would choose `PYODIDE_XBUILDENV_PATH` which is similar to `PYTHONPATH` and `NODE_PATH`.

hoodmane:
Can we by default put it in `xdg_cache_dir/ pyodide_xbuildenvs`?

agriyakhetarpal:
Yes, that would be fine. This option would be in conjunction with that – so that we can place it in a cache dir, but also allow placing it elsewhere if specified.

ryanking13:
> so that we can place it in a cache dir

+1. It would be useful to specify the xbuildenv path so people can cache it in CI.

Issue #72, opened 2024-12-16
Store xbuildenv folders in the xdg-cache directory


agriyakhetarpal:
maybe we should put the xbuildenv folders in the xdg-cache directory. I have so many redundant xbuildenv folders on my computer.

_Originally posted by @hoodmane in https://github.com/pyodide/pyodide/issues/3851#issuecomment-2545473162_

hoodmane:
Thanks!

ryanking13:
Sounds good to me. I think it would be useful to support both a global directory and local directory though, similar to venv or node_modules.

agriyakhetarpal:
Closed with #148; we eventually used [`platformdirs`](https://platformdirs.readthedocs.io/en/stable/) as it turned out that it adheres to [the XDG base directory specification](https://specifications.freedesktop.org/basedir-spec/) already.

PR #148, opened 2025-03-25
Install xbuildenv into the cache directory by default

ryanking13:
As discussed in #72, this changes the xbuildenv installation directory.

Search orders are:

1. user cache directory (using `platformdirs`)
3. pwd

---

@agriyakhetarpal I think you can add `PYODIDE_XBUILDENV_PATH` (#114) as a follow-up.

- [x] changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Added basic support for uv. `uv tool install pyodide-cli --with pyodide-build`, or `uvx --from pyodide-cli --with pyodide-build pyodide --help`, or using `pyodide-build` in `uv`-managed virtual environments will now work.
+  [#132](https://github.com/pyodide/pyodide-build/pull/132)
+
 ### Changed
 
 - The Rust toolchain version has been updated to `nightly-2025-01-18`.
@@ -15,6 +20,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Pyodide cross-build environment will now be installed in the user cache directory by default,
   which is `<home>/.cache` in Linux systems and `/Users/<user>/Library/Caches` in macOS.
   [#148](https://github.com/pyodide/pyodide-build/pull/148)
+
+- Removed `cmake` dependency from the package.
+  The Pyodide build system now uses the `cmake` package specified in the `pyproject.toml` file
+  of the target package.
+  [#141](https://github.com/pyodide/pyodide-build/pull/141)
+
 ### Fixed
 
 - Fixed Pyodide venv `sys_platform` marker evaluation with pip >= 25.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   [#103](https://github.com/pyodide/pyodide-build/pull/103)
 
 - Pyodide cross-build environment will now be installed in the user cache directory by default,
-  which is `<home>/.cache` in unix system and `/Users/<user>/Library/Caches` in MacOS.
+  which is `<home>/.cache` in Linux systems and `/Users/<user>/Library/Caches` in macOS.
   [#148](https://github.com/pyodide/pyodide-build/pull/148)
-
 ### Fixed
 
 - Fixed Pyodide venv `sys_platform` marker evaluation with pip >= 25.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The Rust toolchain version has been updated to `nightly-2025-01-18`.
   [#103](https://github.com/pyodide/pyodide-build/pull/103)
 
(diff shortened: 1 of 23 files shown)


agriyakhetarpal:
Thanks, @ryanking13! I would still insist that we use `platformdirs` here, for three reasons:
- it is not an expensive dependency (no required dependencies of its own)
- we are using `virtualenv`, which depends on it, so `platformdirs` is already being installed right now
- the heaviest dependency in terms of download size is CMake, which will go away in #141, and the heaviest one in terms of the number of required dependencies is `unearth`, for which we have a `TODO:` note to making it optional.

hoodmane:
It feels like typer is the most annoying dependency in terms of how much junk it pulls in.

ryanking13:
> It feels like typer is the most annoying dependency in terms of how much junk it pulls in.

Well, typer itself only has 4 dependencies (click, typing-extensions, shellingham, rich), but `rich` has another 4 dependencies. Though, even if we remove typer, we still use rich directly so we cannot remove it anyways.

pyodide_build/common.py
```diff
@@ -30,11 +32,48 @@
...
-    from pyodide_build import __version__
+    try:
+        from pyodide_build import __version__
+    except ImportError:
+        __version__ = "0.0.0"
 
     return f".pyodide-xbuildenv-{__version__}"
 
 
+@lru_cache(maxsize=1)
```
hoodmane, line 43: I'd suggest using `functools.cache` instead for simplicity.
```suggestion
@cache
```

pyodide_build/common.py
```diff
@@ -30,11 +32,48 @@
...
+    # 1. default cache directory
+    candidates.append(Path(platformdirs.user_cache_dir()) / dirname)
+
+    # 2. current working directory
+    candidates.append(Path.cwd() / dirname)
+
+    for candidate in candidates:
+        if _has_write_access(candidate):
+            return candidate
+
```
hoodmane, line 62: Maybe raise if we fall through to here?

pyodide_build/common.py
```diff
@@ -30,11 +32,48 @@
...
+def _has_write_access(folder: Path) -> bool:
+    """
+    Checks if the current user has write access to the given folder using pathlib.
+    """
+    try:
+        if not folder.exists() and folder.parent != folder:
+            return _has_write_access(folder.parent)
+
+        return os.access(str(folder), os.W_OK)
```
hoodmane, line 72: Should we check for `W_OK | X_OK`? As far as I understand, we can't add a file to a folder without execute permissions?

hoodmane (approved):
LGTM other than a couple of nits. Thanks @ryanking13!

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The Rust toolchain version has been updated to `nightly-2025-01-18`.
   [#103](https://github.com/pyodide/pyodide-build/pull/103)
 
+- Pyodide cross-build environment will now be installed in the user cache directory by default,
+  which is `<home>/.cache` in unix system and `/Users/<user>/Library/Caches` in MacOS.
+  [#148](https://github.com/pyodide/pyodide-build/pull/148)
+
```
agriyakhetarpal, line 18: (nit): macOS is also a Unix-based OS, so:

```suggestion
- Pyodide cross-build environment will now be installed in the user cache directory by default,
  which is `<home>/.cache` in Linux systems and `/Users/<user>/Library/Caches` in macOS.
  [#148](https://github.com/pyodide/pyodide-build/pull/148)
```

pyodide_build/common.py
```diff
@@ -16,12 +16,14 @@
 from collections import deque
 from collections.abc import Generator, Iterable, Iterator, Mapping, Sequence
 from contextlib import contextmanager
+from functools import lru_cache
```
agriyakhetarpal, line 19:
```suggestion
from functools import cache
```

From @hoodmane's suggestion on using `@functools.cache()` above.

pyodide_build/common.py
```diff
@@ -30,11 +32,48 @@
 
 
 def xbuildenv_dirname() -> str:
-    from pyodide_build import __version__
+    try:
+        from pyodide_build import __version__
+    except ImportError:
+        __version__ = "0.0.0"
```
agriyakhetarpal, line 38: should we also check for `ModuleNotFoundError`?
```suggestion
    try:
        from pyodide_build import __version__
    except (ImportError, ModuleNotFoundError):
        __version__ = "0.0.0"
```

pyodide_build/common.py
agriyakhetarpal: > Should we check for `W_OK | X_OK`? As far as I understand, we can't add a file to a folder without execute permissions?

I think `W_OK` is fine. What we should check for are (assuming `Path` is the absolute/resolved path to the directory):

- if `Path.owner() == os.getlogin()` returns `True` (checks for the owner by name); and
- if `Path.stat().st_uid == os.geteuid()` returns `True` (checks for the effective user ID).

These might not be available everywhere (https://docs.python.org/3/library/os.html#os.geteuid), but no one will run `pyodide-build` in WASI anyway.

agriyakhetarpal (approved):
Thanks, @ryanking13!

ryanking13:
Thanks for the reviews!

Issue #138, opened 2025-03-16
Support buildng multiple outputs from a recipe


agriyakhetarpal:
This feature request is based on a requirement that came up in https://github.com/pyodide/pyodide-build/pull/109 – it would be great if we could also support the conda-forge feature for multiple outputs for builds, which would use the same build environment for building multiple variations of a package.

The proposed implementation could be as follows:
- add an `outputs:` field to the recipe that allows subfields such as `script:`, `about:` `post:`, more `requirements:` (and sub-requirements), etc.
- build in the order they are specified, building the first listed output and proceed to the next one if it passes
    - we can stop if it fails; I don't think we need a --keep-going option or building in parallel at the moment
- the graph builder can keep the same logic, more or less – it's just that one recipe will generate $n$ packages instead of just one

The rationale is that many recipes could benefit from this – for example:
- packages such as NumPy/SciPy/etc. and various Meson-based packages can then use their "install tags" integration with Meson to build WASM wheels for `scipy` and `scipy-tests`, which can be installed together in the same environment. this can be used to control build sizes better than stripping away wheels.
- it allows packages to ship multiple variants of their builds: we could ship NumPy + OpenBLAS and NumPy with Netlib BLAS from the same recipe file, just with different build scripts listed together.
- for libraries: SuiteSparse, for example, can be instructed to build different components separately for packages that need only a specific one.

Reference:
- https://github.com/conda-forge/scipy-feedstock/blob/1e665407f6222093df606d9e27553072b7222fc9/recipe/meta.yaml
- https://github.com/conda-forge/suitesparse-feedstock/blob/d0a371ec49dfb0a9dbbcea5815c0f4c71e87d462/recipe/recipe.yaml

ryanking13:
Thanks for the proposal, @agriyakhetarpal. It sounds like an interesting idea to me. However, I think the format of the `outputs` can be more formalized before we proceed to implementation, for instance, how it will work with the lockfile, etc.

agriyakhetarpal:
Coming back to this from https://github.com/pyodide/pyodide/pull/5563, I think the format will remain largely the same as the current recipes, as we would not like or need the added complexity that comes with conda-forge recipes.

For the lockfile: two outputs would build two different packages with two different names and checksums, which would be placed in the `dist/` folder of the recipe dir, next to `meta.yaml`, so I think we should be fine.

Here is what a sample recipe for NumPy and numpy-tests could look like:

```yaml
package:
  name: numpy
[...]

PR #152, opened 2025-03-29
Add support for building without isolation

### LINKED ISSUE #136: Support `--no-isolation` flag from pypa/build for "unisolated" package builds
This feature request is based on a requirement that came up in https://github.com/pyodide/pyodide-build/pull/109 – building a package without isolation would mean that:
- users install PEP 508 build dependencies independently from a requirements file, an environment file, etc. prior to building the package, so that those dependencies are available
    - this is not easy to handle with `pyodide-build` and Pyodide's recipe format upstream, so we should probably not do this and keep installing the dependencies beforehand.
- we don't create a PEP 517 isolated virtual environment for building if `pyodide build --no-isolation`
    - should we also have a boolean `isolated:` field for the recipes that want to opt in to this?

So, supporting this feature for us would be limited to point 2, similar to `-n` option, but we would deviate from `pypa/build`'s behaviour and would actually be `-nx` (where `-x` stands for skipping build dependency checks – see https://build.pypa.io/en/stable/#python--m-build-options).

A package to be built without isolation would, therefore, go through this process:
- we still install their build-time dependencies into the environment
- at the time of building the package, we pass the equivalent args for the `-nx` flags to `DefaultIsolatedEnv`, or we monkeypatch [`IsolatedEnv`](https://build.pypa.io/en/stable/api.html#build.env.IsolatedEnv) to create a persistent venv for installation that we can reuse.

agriyakhetarpal:
## Description

This PR is the first in a series of PRs aiming to unblock #109 and also towards strides in other areas; such as being able to build a package from one directory twice, see https://github.com/pyodide/pyodide/pull/5031. In particular, this adds two new flags: `--no-isolation`/`-n` and `--skip-dependency-check`/`-x`, to the `pyodide build` CLI. These will allow us not to use an isolated virtual environment when building a wheel for a package if required.

Closes #136

### See also

- https://docs.scipy.org/doc/scipy/building/redistributable_binaries.html
- xref #138

### Some TODO items

- [x] Get "unisolated" builds working
- [x] Get tests passing
- [x] Investigate and fix the issue where the compiler wrappers are ending up in temporary directories
- [x] Gotta add tests!!!
	- [x] Basic tests that see if the isolation is working
	- [x] Integration test with NumPy/SciPy/etc.
- [x] Add a CHANGELOG entry

### DIFF
--- a/integration_tests/Makefile
+++ b/integration_tests/Makefile
@@ -23,6 +23,19 @@ test-src: check
 
 	@echo "... Passed"
 
+.PHONY: test-src-no-isolation
+test-src-no-isolation: check
+	@echo "... Running integration tests for building src with --no-isolation --skip-dependency-check"
+
+	# Ensure the necessary build dependencies are installed in the
+	# host environment as we need to do so ourselves.
+
+	pip install meson-python meson cython
+
+	./src/numpy.sh --no-isolation --skip-dependency-check
+
+	@echo "... Passed"
+
 .PHONY: check
 check:
 	@echo "... Checking dependencies"
--- a/integration_tests/Makefile
+++ b/integration_tests/Makefile
@@ -32,7 +32,7 @@ test-src-no-isolation: check
 	@( \
         set -e; \
 		python -m venv ./test_venv; \
-		source ./test_venv/bin/activate; \
+		. ./test_venv/bin/activate; \
 		pip install meson-python meson cython; \
 		pip install -e ../; \
 		./src/numpy_no_isolation.sh; \
--- a/integration_tests/Makefile
+++ b/integration_tests/Makefile
@@ -27,15 +27,21 @@ test-src: check
 test-src-no-isolation: check
 	@echo "... Running integration tests for building src with --no-isolation --skip-dependency-check"
 
-	# Ensure the necessary build dependencies are installed in the
-	# host environment as we need to do so ourselves.
+	# Some virtualenv workarounds from https://stackoverflow.com/a/24736236
+	# to make sure that we are using the right environment
+	@( \
+        set -e; \
+		python -m venv ./test_venv; \
+		source ./test_venv/bin/activate; \
+		pip install meson-python meson cython; \
+		pip install -e ../; \
+		./src/numpy_no_isolation.sh; \
+		deactivate; \
+	)
 
-	pip install meson-python meson cython
-
-	./src/numpy.sh --no-isolation --skip-dependency-check
+	@rm -rf ./test_venv
 
 	@echo "... Passed"
-
 .PHONY: check
 check:
 	@echo "... Checking dependencies"
@@ -49,5 +55,9 @@ check:
 clean:
 	rm -rf .pyodide-xbuildenv*
 	rm -rf recipes/*/build
+	rm -rf test_venv
 	rm -rf src/numpy-*
+	rm -rf src/numpy-*.tar.gz
+	rm -rf numpy-*
+	rm -rf numpy-*.tar.gz
 	rm -rf dist
--- a/integration_tests/src/numpy.sh
+++ b/integration_tests/src/numpy.sh
@@ -2,17 +2,6 @@
 
 set -e
 
-NO_ISOLATION=""
-SKIP_DEPS=""
-
-while [[ "$#" -gt 0 ]]; do
-    case $1 in
-        --no-isolation) NO_ISOLATION="--no-isolation"; shift ;;
-        --skip-dependency-check) SKIP_DEPS="--skip-dependency-check"; shift ;;
-        *) echo "Unknown parameter: $1"; exit 1 ;;
-    esac
-done
-
 VERSION="2.0.2"
 URL="https://files.pythonhosted.org/packages/source/n/numpy/numpy-${VERSION}.tar.gz"
 
@@ -21,4 +10,4 @@ tar -xf numpy-${VERSION}.tar.gz
 cd numpy-${VERSION}
 
 MESON_CROSS_FILE=$(pyodide config get meson_cross_file)
-${UV_RUN_PREFIX} pyodide build -Csetup-args=-Dallow-noblas=true -Csetup-args=--cross-file="${MESON_CROSS_FILE}" $NO_ISOLATION $SKIP_DEPS
+${UV_RUN_PREFIX} pyodide build -Csetup-args=-Dallow-noblas=true -Csetup-args=--cross-file="${MESON_CROSS_FILE}" "$@"
--- a/integration_tests/src/numpy.sh
+++ b/integration_tests/src/numpy.sh
@@ -10,4 +10,4 @@ tar -xf numpy-${VERSION}.tar.gz
 cd numpy-${VERSION}
 
 MESON_CROSS_FILE=$(pyodide config get meson_cross_file)
-${UV_RUN_PREFIX} pyodide build -Csetup-args=-Dallow-noblas=true -Csetup-args=--cross-file="${MESON_CROSS_FILE}" "$@"
+${UV_RUN_PREFIX} pyodide build -Csetup-args=-Dallow-noblas=true -Csetup-args=--cross-file="${MESON_CROSS_FILE}"
--- a/integration_tests/src/numpy.sh
+++ b/integration_tests/src/numpy.sh
@@ -2,6 +2,17 @@
 
 set -e
 
+NO_ISOLATION=""
+SKIP_DEPS=""
+
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        --no-isolation) NO_ISOLATION="--no-isolation"; shift ;;
+        --skip-dependency-check) SKIP_DEPS="--skip-dependency-check"; shift ;;
+        *) echo "Unknown parameter: $1"; exit 1 ;;
+    esac
+done
+
 VERSION="2.0.2"
 URL="https://files.pythonhosted.org/packages/source/n/numpy/numpy-${VERSION}.tar.gz"
 
@@ -10,4 +21,4 @@ tar -xf numpy-${VERSION}.tar.gz
 cd numpy-${VERSION}
 
 MESON_CROSS_FILE=$(pyodide config get meson_cross_file)
-${UV_RUN_PREFIX} pyodide build -Csetup-args=-Dallow-noblas=true -Csetup-args=--cross-file="${MESON_CROSS_FILE}"
+${UV_RUN_PREFIX} pyodide build -Csetup-args=-Dallow-noblas=true -Csetup-args=--cross-file="${MESON_CROSS_FILE}" $NO_ISOLATION $SKIP_DEPS
--- a/integration_tests/src/numpy_no_isolation.sh
+++ b/integration_tests/src/numpy_no_isolation.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 
-# The same as "numpy.sh", but without the isolation.
+# The same as "numpy.sh", but without the isolation, and
+# builds both NumPy and numpy-tests from a persistent
+# build directory.
 
 set -e
 
--- a/integration_tests/src/numpy_no_isolation.sh
+++ b/integration_tests/src/numpy_no_isolation.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+# The same as "numpy.sh", but without the isolation.
+
+set -e
+
+VERSION="2.0.2"
+URL="https://files.pythonhosted.org/packages/source/n/numpy/numpy-${VERSION}.tar.gz"
+
+wget $URL
+tar -xf numpy-${VERSION}.tar.gz
+cd numpy-${VERSION}
+
+MESON_CROSS_FILE=$(pyodide config get meson_cross_file)
+
+# Build in a persistent build directory
+${UV_RUN_PREFIX} pyodide build \
+    -Csetup-args=-Dallow-noblas=true \
+    -Csetup-args=--cross-file="${MESON_CROSS_FILE}" \
+    -Cinstall-args=--tags=runtime,python-runtime,devel \
+    -Cbuild-dir="build" \
+    --no-isolation --skip-dependency-check
+
+sed -i 's/numpy/numpy-tests/g' pyproject.toml
+
+${UV_RUN_PREFIX} pyodide build \
+    -Csetup-args=-Dallow-noblas=true \
+    -Csetup-args=--cross-file="${MESON_CROSS_FILE}" \
+    -Cinstall-args=--tags=tests \
+    -Cbuild-dir="build" \
+    --no-isolation --skip-dependency-check
+
+echo "Successfully built wheels for numpy-${VERSION} and numpy-tests-${VERSION}."
--- a/pyodide_build/cli/build.py
+++ b/pyodide_build/cli/build.py
@@ -46,6 +46,8 @@ def pypi(
     output_directory: Path,
     exports: str,
     config_settings: ConfigSettingsType,
+    isolation: bool = True,
+    skip_dependency_check: bool = False,
 ) -> Path:
     """Fetch a wheel from pypi, or build from source if none available."""
     with tempfile.TemporaryDirectory() as tmpdir:
@@ -65,6 +67,8 @@ def pypi(
             output_directory,
             convert_exports(exports),
             config_settings,
+            isolation=isolation,
+            skip_dependency_check=skip_dependency_check,
         )
         return built_wheel
 
@@ -86,6 +90,8 @@ def url(
     output_directory: Path,
     exports: str,
     config_settings: ConfigSettingsType,
+    isolation: bool = True,
+    skip_dependency_check: bool = False,
 ) -> Path:
     """Fetch a wheel or build sdist from url."""
     with tempfile.TemporaryDirectory() as tmpdir:
@@ -102,7 +108,12 @@ def url(
             # unzipped into subfolder
             builddir = files[0]
         wheel_path = build.run(
-            builddir, output_directory, convert_exports(exports), config_settings
+            builddir,
+            output_directory,
+            convert_exports(exports),
+            config_settings,
+            isolation=isolation,
+            skip_dependency_check=skip_dependency_check,
         )
         return wheel_path
 
@@ -112,10 +123,17 @@ def source(
     output_directory: Path,
     exports: str,
     config_settings: ConfigSettingsType,
+    isolation: bool = True,
+    skip_dependency_check: bool = False,
 ) -> Path:
     """Use pypa/build to build a Python package from source"""
     built_wheel = build.run(
-        source_location, output_directory, convert_exports(exports), config_settings
+        source_location,
+        output_directory,
+        convert_exports(exports),
+        config_settings,
+        isolation=isolation,
+        skip_dependency_check=skip_dependency_check,
     )
     return built_wheel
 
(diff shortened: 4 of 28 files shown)


agriyakhetarpal:
Hi @ryanking13, I've been able to build and test out the feature(s) I've added in this PR, but one problem still exists: the compiler wrappers for Emscripten are still being placed inside temporary directories as a part of the build process for some reason: https://github.com/pyodide/pyodide-build/actions/runs/14151180877/job/39644562112?pr=152#step:11:449

What this means in action is that if I were to build NumPy without isolation via the below command invocation in a persistent build directory, say, in `build/` inside the NumPy repo root:

```bash
pyodide build . \
-Csetup-args="--cross-file=$(pyodide config get meson_cross_file)" \
-Csetup-args="-Dblas=none" -Csetup-args="-Dlapack=none" \
-nx \ # i.e., --no-isolation and --skip-dependency-check args, which this PR adds
-Cbuild-dir=build`
```

I am able to build a NumPy wheel. However, if I were to rerun the same command, hoping to make use of the caching provided by the persistent build directory and achieve a speedup for the subsequent run(s) , it would fail because Meson stores its build metadata inside the build directory, which also contains information about the compiler, etc. Meson then fails to find the compiler, as the temporary directory used the previous wheel build where the compiler wrappers were stored no longer exists.

When I build NumPy with a persistent build dir with native compilation on my macOS machine, the compilers are picked up correctly from conda-forge each time – this makes me think that this is a `pyodide-build` specific issue. Maybe it has something to do with `pywasmcross`?

I think you would be the best person to know why this is the case and on how we can address it, so I thought I should ask you about it! :)

integration_tests/Makefile
```diff
@@ -23,6 +23,19 @@ test-src: check
...
 	@echo "... Passed"
 
+.PHONY: test-src-no-isolation
+test-src-no-isolation: check
+	@echo "... Running integration tests for building src with --no-isolation --skip-dependency-check"
+
+	# Ensure the necessary build dependencies are installed in the
+	# host environment as we need to do so ourselves.
+
+	pip install meson-python meson cython
```
ryanking13, line 33: How about installing these dependencies in a virtualenv to prevent polluting the host environment?

integration_tests/src/numpy.sh
```diff
@@ -2,6 +2,17 @@
...
+NO_ISOLATION=""
+SKIP_DEPS=""
+
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        --no-isolation) NO_ISOLATION="--no-isolation"; shift ;;
+        --skip-dependency-check) SKIP_DEPS="--skip-dependency-check"; shift ;;
+        *) echo "Unknown parameter: $1"; exit 1 ;;
+    esac
+done
```
ryanking13, line 14: Maybe we can just forward `$@` to `pyodide build`, instead of handling these parameters separately?

ryanking13:
Hi @agriyakhetarpal, thanks for working on this.

> the compiler wrappers for Emscripten are still being placed inside temporary directories as a part of the build process for some reason:

Yep, we are using temporary directory for now, but I think you can change it. We've already changed it in recipe build to store the symlinks in the build directory (https://github.com/pyodide/pyodide-build/pull/96), so I think you can do the similar thing in the `pyodide build`.

The directory where the symlinks are stored is decided [here](https://github.com/pyodide/pyodide-build/blob/71aa0b5ae55b196c4b3444434c67a9541dd382c9/pyodide_build/pypabuild.py#L302-L303) and [here](https://github.com/pyodide/pyodide-build/blob/71aa0b5ae55b196c4b3444434c67a9541dd382c9/pyodide_build/pypabuild.py#L257-L265). I think you can pass the `build_dir` parameter [here](https://github.com/pyodide/pyodide-build/blob/71aa0b5ae55b196c4b3444434c67a9541dd382c9/pyodide_build/out_of_tree/build.py#L31-L39) to store the symlink.

pyodide_build/pypabuild.py
```diff
@@ -5,6 +5,16 @@
...
 import traceback
 from collections.abc import Callable, Iterator, Mapping, Sequence
+
+
+# A helper function from pypa/build/__main__.py since
+# it's not in the vendorized code we have
+# TODO: we should move this to a new file. it's out of place
+# between the other imports ;-)
+def _format_dep_chain(dep_chain: Sequence[str]) -> str:
+    return " -> ".join(dep.partition(";")[0].strip() for dep in dep_chain)
```
ryanking13, line 15: Maybe you can put it in `common.py`.

We need to reorganize common.py too, as it contains too much code anyway.

pyodide_build/cli/build.py
```diff
@@ -164,6 +182,19 @@ def main(
...
+        "--no-isolation",
+        "-n",
+        help="Disable building the project in an isolated virtual environment. "
+        "Build dependencies must be installed separately when this option is used",
+    ),
+    skip_dependency_check: bool = typer.Option(
+        False,
+        "--skip-dependency-check",
+        "-x",
+        help="Do not check that build dependencies are installed",
```
ryanking13, line 196: Could you please mention that it is no-op when it is not used with `--no-isolation` IIUC?

pyodide_build/pypabuild.py
```diff
@@ -79,7 +89,15 @@ def _runner(cmd, cwd=None, extra_environ=None):
...
         # and we should add it to the PATH so that they can be found.
-        env["BUILD_ENV_SCRIPTS_DIR"] = isolated_build_env.scripts_dir
+        if isolated_build_env:
+            env["BUILD_ENV_SCRIPTS_DIR"] = isolated_build_env.scripts_dir
+        else:
+            # For non-isolated builds, set a fallback path or use the current Python path
+            import sysconfig
+
+            scripts_dir = sysconfig.get_path("scripts")
+            env["BUILD_ENV_SCRIPTS_DIR"] = scripts_dir
```
ryanking13, line 99: I am curious if this is adapted from pypa/build.

pyodide_build/pypabuild.py
```diff
@@ -264,10 +323,22 @@ def _create_symlink_dir(env: dict[str, str], build_dir: Path | None):
...
+    # directory, which breaks persistent builds. This is not ideal, but
+    # it is better than nothing so this is non-blocking for now. It has
+    # to be investigated further.
+
+    if no_isolation:
+        # For non-isolated builds, create a persistent directory in the current working directory
+        # or in a well-known location like ~/.pyodide/compiler_wrappers
+        symlink_dir = Path.cwd() / ".pyodide_compiler_wrappers"
+        symlink_dir.mkdir(exist_ok=True)
+        yield symlink_dir
```
ryanking13, line 336: I don't think we need to branch out this part. I gues we can pass the persistent directory for both no-isolated and isolated build (as I commented in https://github.com/pyodide/pyodide-build/pull/152#issuecomment-2764406264)

pyodide_build/out_of_tree/pypi.py
```diff
@@ -404,7 +418,9 @@ def build_dependencies_for_wheel(
     skip_dependency: list[str],
     exports: _BuildSpecExports,
     config_settings: ConfigSettingsType,
-    output_lockfile: str | None,
+    isolation: bool = True,
+    skip_dependency_check: bool = False,
```
ryanking13, line 422: (No action is required for this PR)

It may be weird to build all the dependencies in a no-isolation mode. But let's go as-is for now. I personally want to deprecate the feature to build the dependencies if package maintainers in Python ecosystem starts to build Pyodide wheel themselves.

pyodide_build/pypabuild.py
```diff
@@ -191,6 +209,45 @@ def _build_in_isolated_env(
...
+) -> str:
+    with common.replace_env(build_env):
+        # Setup sysconfigdata path in environment
+        sysconfigdata_name = get_build_flag("SYSCONFIG_NAME")
+        sysconfig_dir = Path(get_build_flag("TARGETINSTALLDIR")) / "sysconfigdata"
+        env = os.environ.copy()
+        env["_PYTHON_SYSCONFIGDATA_NAME"] = sysconfigdata_name
+        env["PYTHONPATH"] = f"{str(sysconfig_dir)}:{env.get('PYTHONPATH', '')}"
+
+        with common.replace_env(env):
```
ryanking13, line 228: I guess `build_env` created from `get_build_env` would already have these (`_PYTHON_SYSCONFIGDATA_NAME`, `PYTHONPATH`) values in the map. Why does this need to re-calculated?

pyodide_build/pypabuild.py
```diff
@@ -191,6 +209,45 @@ def _build_in_isolated_env(
...
+            builder = _ProjectBuilder(srcdir, runner=_gen_runner(build_env))
+
+            if not skip_dependency_check:
+                missing = builder.check_dependencies(
```
ryanking13, line 232: 👍 

[`check_dependencies`](https://github.com/pypa/build/blob/811c9db0dcac14b7b32f08ee57d09af3666979d4/src/build/_builder.py#L242) looks like a union of

1. get_requires_for_build (== dependencies for the build backend such as setuptools, meson)
2. build_system_requires (== dependencies in the pyproject.toml)

In the isolated build, we've been [dealing with them separately](https://github.com/pyodide/pyodide-build/blob/71aa0b5ae55b196c4b3444434c67a9541dd382c9/pyodide_build/pypabuild.py#L160-L167). I guess we can use `check_dependencies` there too (no action required for this PR).

pyodide_build/pypabuild.py
```diff
@@ -191,6 +209,45 @@ def _build_in_isolated_env(
...
+                    distribution, config_settings or {}
+                )
+                if missing:
+                    dependencies = "".join(
+                        "\n\t" + dep
+                        for deps in missing
+                        for dep in (deps[0], _format_dep_chain(deps[1:]))
+                        if dep
+                    )
+                    _error(f"Missing dependencies:{dependencies}")
```
ryanking13, line 242: How about putting all these in a single function?

pyodide_build/out_of_tree/pypi.py
hoodmane: Joe Marshall added this and we were always a bit unsure if it made sense as a feature.

agriyakhetarpal: Yes, I added these extra args to normalise what passes through these functions, even if unused. I agree that it makes sense to drop it entirely. I don't think I've used it personally, and I I haven't seen it being used in the main Pyodide repo.

pyodide_build/pypabuild.py
agriyakhetarpal: > Why does this need to re-calculated?

It doesn't; this was an oversight. Thanks for the catch!

pyodide_build/pypabuild.py
agriyakhetarpal: Yes, sort of. `pypa/build` has a more robust check for where the `bin/` dir is usually placed. It's here: https://github.com/pypa/build/blob/811c9db0dcac14b7b32f08ee57d09af3666979d4/src/build/env.py#L336-L376. I used a subset of it through `sysconfig.get_path("scripts")`, as we don't need to support all sorts of Python distributions that might be packaged differently.

pyodide_build/pypabuild.py
agriyakhetarpal: Addressed in 8530171ac2f30bbb18099089382cffb89ca7c51d.

pyodide_build/pypabuild.py
agriyakhetarpal: Thanks for the pointer. I addressed this in 8325e803d78cc3e3833edceeb652aee2e7974891.

pyodide_build/cli/build.py
agriyakhetarpal: Addressed in 6c5da50434a00862e3e7d732504d62780fc6c3f8.

pyodide_build/pypabuild.py
agriyakhetarpal: I agree. For the time being, I addressed this in 0aed06b6b02ba02dd08d9686a7c00d4e54437129.

integration_tests/src/numpy.sh
agriyakhetarpal: Addressed in e38a704953522fcdc7366f612d52b4c4e77e798b.

integration_tests/Makefile
agriyakhetarpal: 🤔 I think I can't activate a virtual environment inside a Makefile for some reason. Maybe that's not possible. I'll look for a workaround.

agriyakhetarpal:
> Yep, we are using temporary directory for now, but I think you can change it. We've already changed it in recipe build to store the symlinks in the build directory (#96), so I think you can do the similar thing in the `pyodide build`.
> 
> The directory where the symlinks are stored is decided [here](https://github.com/pyodide/pyodide-build/blob/71aa0b5ae55b196c4b3444434c67a9541dd382c9/pyodide_build/pypabuild.py#L302-L303) and [here](https://github.com/pyodide/pyodide-build/blob/71aa0b5ae55b196c4b3444434c67a9541dd382c9/pyodide_build/pypabuild.py#L257-L265). I think you can pass the `build_dir` parameter [here](https://github.com/pyodide/pyodide-build/blob/71aa0b5ae55b196c4b3444434c67a9541dd382c9/pyodide_build/out_of_tree/build.py#L31-L39) to store the symlink.

Thanks a lot for the pointer! I'm able to get subsequent builds for NumPy working. :D

I was able to compile NumPy in 77 seconds on my machine, and taking advantage of the build caching through the persistence of the build dir, `numpy-tests` compiles in 8 seconds.

integration_tests/src/numpy.sh
```diff
@@ -10,4 +10,4 @@ tar -xf numpy-${VERSION}.tar.gz
 cd numpy-${VERSION}
 
 MESON_CROSS_FILE=$(pyodide config get meson_cross_file)
-${UV_RUN_PREFIX} pyodide build -Csetup-args=-Dallow-noblas=true -Csetup-args=--cross-file="${MESON_CROSS_FILE}"
+${UV_RUN_PREFIX} pyodide build -Csetup-args=-Dallow-noblas=true -Csetup-args=--cross-file="${MESON_CROSS_FILE}" "$@"
```
agriyakhetarpal, line 13:
```suggestion
${UV_RUN_PREFIX} pyodide build -Csetup-args=-Dallow-noblas=true -Csetup-args=--cross-file="${MESON_CROSS_FILE}"
```

integration_tests/src/numpy_no_isolation.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+# The same as "numpy.sh", but without the isolation.
```
agriyakhetarpal, line 3:
```suggestion
# The same as "numpy.sh", but without the isolation, and
# builds both NumPy and numpy-tests from a persistent
# build directory.
```

ryanking13:
Thanks again Agriya! I didn't forget this PR, but I've been a little busy this week and haven't been keeping up with all the PRs. I would like to review this in detail, so please wait for me until the weekend! 🙏

agriyakhetarpal:
No rush, thanks for the update!

agriyakhetarpal:
No rush, thank you! :)

pyodide_build/common.py
```diff
@@ -383,6 +383,19 @@ def _get_sha256_checksum(archive: Path) -> str:
     return h.hexdigest()
 
 
+def _format_dep_chain(dep_chain: Sequence[str]) -> str:
+    return " -> ".join(dep.partition(";")[0].strip() for dep in dep_chain)
+
+
+def _format_missing_dependencies(missing) -> str:
```
ryanking13, line 390: Could you please add a type for the parameter?

pyodide_build/pywasmcross.py
```diff
@@ -521,10 +521,6 @@ def handle_command_generate_args(  # noqa: C901
             "emcmake",
             "cmake",
             *flags,
-            # Since we create a temporary directory and install compiler symlinks every time,
```
ryanking13, line 524: Oh, yes, we can remove this now. 👍

pyodide_build/out_of_tree/build.py
```diff
@@ -28,6 +31,19 @@ def run(
...
+    build_dir.mkdir(exist_ok=True)
+    # don't track the build dir (helps if building in a git repo)
+    gitignore_path = build_dir / ".gitignore"
+    if not gitignore_path.exists():
+        with gitignore_path.open("w") as f:
+            f.write(
+                dedent("""\
+                    # Created by pyodide-build
+                    *""")
+            )
```
ryanking13, line 45: Could you extract this out to a function, such as `prepare_build_dir(build_dir)`?

pyodide_build/pypabuild.py
ryanking13: Makes sense. Thanks for the explanation!

pyodide_build/pypabuild.py
```diff
@@ -254,20 +285,14 @@ def make_command_wrapper_symlinks(symlink_dir: Path) -> dict[str, str]:
...
-        symlink_dir.mkdir()
-        yield symlink_dir
-        return
-
-    # Running from "pyodide build". Put symlinks in a temporary directory.
-    # TODO: Add a debug option to save the symlinks.
-    with TemporaryDirectory() as symlink_dir_str:
-        yield Path(symlink_dir_str)
+def _create_symlink_dir(
+    env: dict[str, str], build_dir: Path, no_isolation: bool = False
```
ryanking13, line 289: no_isolation parameter is not used anymore

pyodide_build/pypabuild.py
```diff
@@ -254,20 +285,14 @@ def make_command_wrapper_symlinks(symlink_dir: Path) -> dict[str, str]:
...
-        shutil.rmtree(symlink_dir, ignore_errors=True)
-        symlink_dir.mkdir()
-        yield symlink_dir
-        return
-
-    # Running from "pyodide build". Put symlinks in a temporary directory.
-    # TODO: Add a debug option to save the symlinks.
-    with TemporaryDirectory() as symlink_dir_str:
-        yield Path(symlink_dir_str)
+def _create_symlink_dir(
```
ryanking13, line 288: We are not using a temporary directory anymore, so I don't think this needs to be a context manager anymore. No cleanup is done at the function end.

ryanking13 (approved):
I left some minor comments, but it looks good to me overall.

> I was able to compile NumPy in 77 seconds on my machine, and taking advantage of the build caching through the persistence of the build dir, numpy-tests compiles in 8 seconds.

Nice!

BTW, some of the Google's Bazel-based packages were relying on no-isolated builds, such as JAX. There were a lot of other build issues as well, but I think would make one step closer to supporting those packages. Thank you!

pyodide_build/common.py
agriyakhetarpal: Done, thanks! Addressed in 206a28e68c8b731827e7876048beb90f16179607

agriyakhetarpal: Done, thanks! Addressed in 206a28e68c8b731827e7876048beb90f16179607.

pyodide_build/out_of_tree/build.py
agriyakhetarpal: Yes, this makes sense – thanks. I extracted it out to `_prepare_build_dir` and also added a VCS-ignore file for Mercurial in de93d2b62c6119ea245dbeeb674d0b0301f20bca and 6ef79be4e11702e9139af8001b989859548ea2d8 respectively.

pyodide_build/pypabuild.py
agriyakhetarpal: Good catch – dropped in 920aec7a56a778df4bf25961ebff9056f1655803.

pyodide_build/pypabuild.py
agriyakhetarpal: Moved this to a new issue: #175.

pyodide_build/pypabuild.py
agriyakhetarpal: Moved it to a new issue: #176

agriyakhetarpal:
Many thanks for the detailed review, @ryanking13! I'll merge this now, and I've opened issues for adding follow-ups.

PR #154, opened 2025-03-29
Better check for write access when determining default xbuildenv path

agriyakhetarpal:
## Description

As a small follow-up to the changes made in #148, this PR adds a better check to evaluate if we have write access to the xbuildenv directory. It also adds the relevant attribution to the code that I had forgotten to attach in the review there.

### DIFF
--- a/pyodide_build/common.py
+++ b/pyodide_build/common.py
@@ -75,10 +75,11 @@ def _has_write_access(folder: Path) -> bool:
     Checks if the current user has write access to the given folder using pathlib.
     """
     try:
-        p = folder.resolve()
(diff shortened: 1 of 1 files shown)


agriyakhetarpal:
Hi Gyeongjae, I hope it is okay with you that I've addressed my previous review comment in https://github.com/pyodide/pyodide-build/pull/148/files#r2016356901 here. Thanks!

ryanking13 (approved):
Sounds good. Thanks!

agriyakhetarpal:
Thanks for the review!

PR #158, opened 2025-03-31
Support custom xbuildenv paths, and add xbuildenv path to `ConfigManager`

### LINKED ISSUE #114: Support `--xbuildenv-path` flag and `PYODIDE_XBUILDENV_PATH` environment variable with `pyodide build`

agriyakhetarpal:
## Description

This PR closes #114, and is a follow-up to #148 and #154. adds the following:

1. an `--xbuildenv-path` argument to the `pyodide build` CLI that gets passed on to `pyodide xbuildenv install --path`.
2. an `xbuildenv_path` config variable to the `pyodide config` list, so that people can configure a default path under the `[tool.pyodide.build]` table in `pyproject.toml`
3. a `PYODIDE_XBUILDENV_PATH` environment variable that can be used for configuration.

The order of precedence is:
- `PYODIDE_XBUILDENV_PATH`, then
- the value of the `pyodide config get xbuildenv_path` command (supports relative paths as well)
- the default location, i.e., what is defined by `platformdirs` across Linux/macOS devices

Use cases:
- set a persistent xbuildenv path at a fixed location of the user's choice, instead of what is set by `platformdirs`
- use the environment variable to cache the xbuildenv in CI
- and so on

### DIFF
--- a/pyodide_build/cli/config.py
+++ b/pyodide_build/cli/config.py
@@ -34,9 +34,6 @@ def _get_configs() -> dict[str, str]:
 
     configs: dict[str, str] = get_build_environment_vars(get_pyodide_root())
 
-    # Print out empty config variables as well, instead of ignoring them. It
-    # helps with visibility as to what is being used in the builds, and what
-    # isn't being used.
     configs_filtered = {k: configs.get(v, "") for k, v in PYODIDE_CONFIGS.items()}
 
     return configs_filtered
--- a/pyodide_build/cli/config.py
+++ b/pyodide_build/cli/config.py
@@ -20,6 +20,7 @@ PYODIDE_CONFIGS = {
     "cxxflags": "SIDE_MODULE_CXXFLAGS",
     "ldflags": "SIDE_MODULE_LDFLAGS",
     "meson_cross_file": "MESON_CROSS_FILE",
+    "xbuildenv_path": "PYODIDE_XBUILDENV_PATH",
 }
 
 
@@ -33,7 +34,11 @@ def _get_configs() -> dict[str, str]:
 
     configs: dict[str, str] = get_build_environment_vars(get_pyodide_root())
 
-    configs_filtered = {k: configs[v] for k, v in PYODIDE_CONFIGS.items()}
+    # Print out empty config variables as well, instead of ignoring them. It
+    # helps with visibility as to what is being used in the builds, and what
+    # isn't being used.
+    configs_filtered = {k: configs.get(v, "") for k, v in PYODIDE_CONFIGS.items()}
+
     return configs_filtered
 
 
--- a/pyodide_build/cli/config.py
+++ b/pyodide_build/cli/config.py
@@ -34,8 +34,7 @@ def _get_configs() -> dict[str, str]:
 
     configs: dict[str, str] = get_build_environment_vars(get_pyodide_root())
 
-    configs_filtered = {k: configs.get(v, "") for k, v in PYODIDE_CONFIGS.items()}
-
+    configs_filtered = {k: configs[v] for k, v in PYODIDE_CONFIGS.items()}
     return configs_filtered
 
 
--- a/pyodide_build/cli/xbuildenv.py
+++ b/pyodide_build/cli/xbuildenv.py
@@ -35,7 +35,9 @@ def _install(
         None, help="version of cross-build environment to install"
     ),
     path: Path = typer.Option(
-        DEFAULT_PATH, help="path to cross-build environment directory"
+        DEFAULT_PATH,
+        envvar="PYODIDE_XBUILDENV_PATH",
+        help="path to download cross-build environment directory to.",
     ),
     url: str = typer.Option(None, help="URL to download cross-build environment from"),
     force_install: bool = typer.Option(
--- a/pyodide_build/cli/xbuildenv.py
+++ b/pyodide_build/cli/xbuildenv.py
@@ -37,7 +37,7 @@ def _install(
     path: Path = typer.Option(
         DEFAULT_PATH,
         envvar="PYODIDE_XBUILDENV_PATH",
-        help="path to download cross-build environment directory to.",
+        help="destination to download cross-build environment directory to.",
     ),
     url: str = typer.Option(None, help="URL to download cross-build environment from"),
     force_install: bool = typer.Option(
--- a/pyodide_build/common.py
+++ b/pyodide_build/common.py
@@ -64,7 +64,7 @@ def default_xbuildenv_path() -> Path:
         if _has_write_access(path):
             return path
         else:
-            logger.warning(
+            logger.error(
                 "The directory specified in PYODIDE_XBUILDENV_PATH (%s) is not writable. "
                 "Falling back to other locations.",
                 path,
@@ -86,13 +86,13 @@ def default_xbuildenv_path() -> Path:
             if _has_write_access(config_path):
                 return config_path
             else:
-                warning_msg = (
-                    f"The directory specified in pyproject.toml ({config_path}) is not writable. "
-                    "Falling back to default locations."
+                logger.error(
+                    "The directory specified in pyproject.toml (%s) is not writable. "
+                    "Falling back to default locations.",
+                    config_path,
                 )
-                logger.warning(warning_msg)
     except Exception as e:
-        logger.warning("Error reading xbuildenv_path from config system: %s", e)
+        logger.error("Error reading xbuildenv_path from config system: %s", e)
 
     # 3. use default locations from platformdirs and elsewhere
     dirname = xbuildenv_dirname()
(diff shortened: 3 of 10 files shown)


pyodide_build/cli/xbuildenv.py
```diff
@@ -35,7 +35,9 @@ def _install(
         None, help="version of cross-build environment to install"
     ),
     path: Path = typer.Option(
-        DEFAULT_PATH, help="path to cross-build environment directory"
+        DEFAULT_PATH,
+        envvar="PYODIDE_XBUILDENV_PATH",
+        help="path to download cross-build environment directory to.",
```
hoodmane, line 40: Maybe:
```suggestion
        help="destination to download cross-build environment directory to.",
```

pyodide_build/common.py
```diff
@@ -43,17 +43,64 @@ def xbuildenv_dirname() -> str:
...
+    if env_path:
+        path = Path(env_path)
+        if _has_write_access(path):
+            return path
+        else:
+            logger.warning(
+                "The directory specified in PYODIDE_XBUILDENV_PATH (%s) is not writable. "
+                "Falling back to other locations.",
+                path,
+            )
```
hoodmane, line 71: Maybe we should fail here?

agriyakhetarpal: I was thinking of it as well and was unsure, thanks for raising it – will change the warnings to errors.

agriyakhetarpal: Addressed by c83f0ae3549f206fcefe02b161409d11f28f6de0.

pyodide_build/cli/config.py
```diff
@@ -33,7 +34,8 @@ def _get_configs() -> dict[str, str]:
 
     configs: dict[str, str] = get_build_environment_vars(get_pyodide_root())
 
-    configs_filtered = {k: configs[v] for k, v in PYODIDE_CONFIGS.items()}
+    configs_filtered = {k: configs.get(v, "") for k, v in PYODIDE_CONFIGS.items()}
```
ryanking13, line 37: I think PYODIDE_CONFIGS should be strictly a subset of the return value of `get_build_environment_vars`. Showing the empty value that does not exist would not be very useful.

pyodide_build/common.py
```diff
@@ -43,17 +43,64 @@ def xbuildenv_dirname() -> str:
...
+        else:
+            logger.error(
+                "The directory specified in PYODIDE_XBUILDENV_PATH (%s) is not writable. "
+                "Falling back to other locations.",
+                path,
+            )
+
+    # 2. check "pyodide config xbuildenv_path"
+    try:
+        from pyodide_build.config import ConfigManager
```
ryanking13, line 75: You can use [`get_host_build_flag("PYODIDE_XBUILDENV_PATH")`](https://github.com/pyodide/pyodide-build/blob/c83f0ae3549f206fcefe02b161409d11f28f6de0/pyodide_build/build_env.py#L166-L174) instead of instantiating the ConfigManager.

Also, the env variable [takes dominance](https://github.com/pyodide/pyodide-build/blob/c83f0ae3549f206fcefe02b161409d11f28f6de0/pyodide_build/config.py#L26-L32), so I think you don't need (1).

pyodide_build/tests/test_common.py
```diff
@@ -158,32 +161,80 @@ def test_check_wasm_magic_number(tmp_path):
     assert check_wasm_magic_number(tmp_path / "badfile.so") is False
 
 
-def test_default_xbuildenv_path(tmp_path, reset_cache):
-    import platformdirs
+class MockConfigManager:
```
ryanking13, line 164: If `get_host_build_flag` is used (from my previous comment), we won't need to the mock config manager anymore. We can mock the return value of `get_host_build_flag` instead.

Also, the functionality of the config manager is tested separately in `test_config.py`, so I don't think we need multiple cases to test the precedence of env var and config file here.

ryanking13 (commented):
Thanks for working on this!

pyodide_build/cli/config.py
agriyakhetarpal: Yes, that is fair. I dropped this change in 8280f05956533c7200fda6090431d7f3e463c124

pyodide_build/common.py
agriyakhetarpal: Thanks for the pointer, I didn't know about this! It's much simpler, and I checked – the environment variable does dominate. I addressed this in eefc20c254a4e74adcf4faf9f2af5c9ae45c48b6.

agriyakhetarpal:
I'll mark it as a draft for now until I figure out why the tests break, and also change the tests to not mock the `ConfigManager` class. I think I can just add to the existing tests to `test_config.py` instead.

agriyakhetarpal:
Okay, switching to `get_host_build_flag("PYODIDE_XBUILDENV_PATH")` is what makes them break. I'm not suggesting that your suggestion is wrong, but rather my implementation of it is. I found the difference that causes the bug. :)

In the meantime, I've adjusted a few tests, which are still going to fail – I'll fix them later in the day.

ryanking13:
I think you need to fix the dummy_xbuildenv fixture to use the `PYODIDE_XBUILDENV_PATH` env variable, instead of mocking the platformdirs.

https://github.com/pyodide/pyodide-build/blob/f838f6ac8f6016c85a8a3548dfeb715c1f2adb2b/pyodide_build/tests/conftest.py#L97-L111

agriyakhetarpal:
Thanks, Gyeongjae! The cause turned out to be that `pathlib.Path("")` returns `PosixPath('.')`, so we needed to special-case it. However, I've also addressed your review comment above and mocked `PYODIDE_XBUILDENV_PATH` in that fixture in d8dec09c3c02a318392b3ed0f5ec9af6a6a71de0. This should now be ready for review again.

ryanking13 (approved):
Thanks @agriyakhetarpal!

In my last review, I commented that testing the behavior of ConfigManager is not required. So I understand why you removed all the testcases. But I think one test case for checking if setting `PYODIDE_XBUILDENV_PATH` env variable change the behavior of `default_xbuildenv_path()` would be useful.

agriyakhetarpal:
The failures are because PyPI is having an outage. Installing packages is painfully slow at the moment :/

agriyakhetarpal:
@hoodmane, could you please take a look at the failing integration tests with me here? They seem to appear with https://github.com/pyodide/pyodide-build/pull/144 as they also failed there, so they look unrelated to me. Do you know if it's safe to merge this PR with them failing?

pyodide_build/tests/conftest.py
```diff
@@ -94,17 +94,24 @@ def dummy_xbuildenv_url(httpserver):
...
     assert "PYODIDE_ROOT" not in os.environ
 
-    # use PYODIDE_XBUILDENV_PATH instead patching platformdirs after #114
-    monkeypatch.setattr(platformdirs, "user_cache_dir", lambda: tmp_path)
+    xbuildenv_dir = tmp_path / xbuildenv_dirname()
+
+    def mock_get_host_build_flag(flag_name):
+        if flag_name == "PYODIDE_XBUILDENV_PATH":
+            return str(xbuildenv_dir)
+        return ""
```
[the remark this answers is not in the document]
agriyakhetarpal: That makes sense, addressed in 8d643aee0b4318f52e3dff3372cab3db85d45fa5.

agriyakhetarpal:
> Thanks @agriyakhetarpal!
> 
> In my last review, I commented that testing the behavior of ConfigManager is not required. So I understand why you removed all the testcases. But I think one test case for checking if setting `PYODIDE_XBUILDENV_PATH` env variable change the behavior of `default_xbuildenv_path()` would be useful.

I agree with this; I added one test back which sets the `PYODIDE_XBUILDENV_PATH` environment variable to an assortment of values while creating an xbuildenv, similar to the XDG cache path test.

I also had to revert the changes I made in #154 in this PR, as I found that it didn't work correctly – it was reporting the in `/usr/bin/` directory and subdirectories as writeable when they aren't (of course).

agriyakhetarpal:
Thanks for the reviews, @ryanking13 and @hoodmane!

hoodmane:
Thanks for doing the work! It's great to see our tooling getting more mature.

hoodmane:
This also didn't get a very good commit message.

agriyakhetarpal:
Argh... I removed a bunch of the "Trigger [integration] tests" messages I added, but didn't add the PR description. Sorry for this! I think I should manually add the commit message before I merge.

hoodmane:
> I should manually add the commit message before I merge.

Yeah please do. There's no way of generating satisfactory commit messages without  manually checking.

PR #161, opened 2025-03-31
Fix two problems with pyodide venv

hoodmane:
1. We intended to say `getattr(pkg, "package_type", None) == "shared_library"`
2. If the lock file has a url in the filename field, put the url directly into the simple index.


agriyakhetarpal (approved):
Thanks, @hoodmane!

Issue #163, opened 2025-04-01
"Type not yet supported" when running pyodide-build 0.29.3


iFreilicht:
I just tried to explore the CLI after installing pyodide-build from pypi:

```
pyodide --version
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /Users/feuh/repos/Netlight/Viridium/rr_plus_poc/.venv/bin/pyodide:8 in <module>                  │
│                                                                                                  │
│   5 from pyodide_cli.__main__ import main                                                        │
│   6 if __name__ == "__main__":                                                                   │
│   7 │   sys.argv[0] = re.sub(r"(-script\.pyw|\.exe)?$", "", sys.argv[0])                         │
│ ❱ 8 │   sys.exit(main())                                                                         │
│   9                                                                                              │
│                                                                                                  │
│ /Users/feuh/repos/Netlight/Viridium/rr_plus_poc/.venv/lib/python3.12/site-packages/pyodide_cli/a │
│ pp.py:85 in main                                                                                 │
│                                                                                                  │
│   82                                                                                             │
│   83 def main():                                                                                 │
│   84 │   register_plugins()                                                                      │
│ ❱ 85 │   app()                                                                                   │
│   86                                                                                             │
│   87                                                                                             │
│   88 if "sphinx" in sys.modules and __name__ != "__main__":                                      │
│                                                                                                  │
│ /Users/feuh/repos/Netlight/Viridium/rr_plus_poc/.venv/lib/python3.12/site-packages/typer/main.py │
│ :328 in __call__                                                                                 │
│                                                                                                  │
│ /Users/feuh/repos/Netlight/Viridium/rr_plus_poc/.venv/lib/python3.12/site-packages/typer/main.py │
│ :311 in __call__                                                                                 │
│                                                                                                  │
│ /Users/feuh/repos/Netlight/Viridium/rr_plus_poc/.venv/lib/python3.12/site-packages/typer/main.py │
│ :350 in get_command                                                                              │
│                                                                                                  │
│ /Users/feuh/repos/Netlight/Viridium/rr_plus_poc/.venv/lib/python3.12/site-packages/typer/main.py │
│ :332 in get_group                                                                                │
│                                                                                                  │
│ /Users/feuh/repos/Netlight/Viridium/rr_plus_poc/.venv/lib/python3.12/site-packages/typer/main.py │
│ :491 in get_group_from_info                                                                      │
│                                                                                                  │
[...]

PR #164, opened 2025-04-01
Typer>=0.12.4 supports `Union` types

agriyakhetarpal:
## Description

See #163. Typer now supports union types post https://github.com/fastapi/typer/pull/548. It turns out that we have a lower bound on `typer`, i.e., we already support `typer>0.13`, so this PR just fixes the type annotations.


ryanking13:
That's a good news :)

PR #171, opened 2025-04-04
Update auditwheel-emscripten

hoodmane:


ryanking13 (approved):
Thanks!

hoodmane:
Actually, this will break the stable branch since Emscripten 3.1.58 dynamic libraries are not allowed to have `RUNTIME_PATH` sections. So I think we'll need to add a flag to auditwheel-emscripten and make this opt in.

agriyakhetarpal (approved):

ryanking13:
> Actually, this will break the stable branch since Emscripten 3.1.58 dynamic libraries are not allowed to have RUNTIME_PATH sections. So I think we'll need to add a flag to auditwheel-emscripten and make this opt in.

The behavior of `pyodide auditwheel repair` is changed, but pyodide-build is not affected by it because we are using the API. So, just upgrading the pyodide-build does not add a runtime path yet, I need to update the [corresponding code](https://github.com/pyodide/pyodide-build/blob/f838f6ac8f6016c85a8a3548dfeb715c1f2adb2b/pyodide_build/recipe/builder.py#L715).

hoodmane:
Thanks for the explanation.

PR #173, opened 2025-04-05
Modify runtime path when PYODIDE_ABI_VERSION > 2025

ryanking13:
Follow up work for auditwheel-emscripten update


hoodmane:
Thanks @ryanking13!

hoodmane (approved):

agriyakhetarpal (commented):
Thanks, @ryanking13! I had two questions for my understanding, not blocking the merge:
- Will `vendor-sharedlib` default to `true` at a later point in time?
- Should we relax the `"auditwheel-emscripten~=0.0.9"` requirement for testing these changes?

agriyakhetarpal:
Also, should this be included in the next release that I'm planning?

hoodmane:
No reason not to include it right?

agriyakhetarpal:
Yes, indeed!

ryanking13:
> Will vendor-sharedlib default to true at a later point in time?

I hope so. But, in that case I think we need to un-unvendor the libraries in the `pyodide-lock` to optimize the package loading.

PR #178, opened 2025-04-05
Don't use a context manager for `_create_symlink_dir`

### LINKED ISSUE #176: _create_symlink_dir should no longer be a context manager
Post #152, we are not using a temporary directory in `_create_symlink_dir` anymore, so I don't think it needs to be a context manager anymore. No cleanup is done at the function end.

_Originally posted by @ryanking13 in https://github.com/pyodide/pyodide-build/pull/152#discussion_r2029705103_

agriyakhetarpal:
Potentially closes #176

cc: @ryanking13

### DIFF
--- a/pyodide_build/pypabuild.py
+++ b/pyodide_build/pypabuild.py
@@ -284,17 +284,23 @@ def make_command_wrapper_symlinks(symlink_dir: Path) -> dict[str, str]:
     return env
 
 
-# TODO: a context manager is no longer needed here
-@contextmanager
-def _create_symlink_dir(
-    env: dict[str, str],
-    build_dir: Path,
-):
-    # Leave the symlinks in the build directory. This helps with reproducing.
-    symlink_dir = build_dir / "pywasmcross_symlinks"
-    shutil.rmtree(symlink_dir, ignore_errors=True)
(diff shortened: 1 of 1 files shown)


pyodide_build/pypabuild.py
```diff
@@ -284,17 +284,23 @@ def make_command_wrapper_symlinks(symlink_dir: Path) -> dict[str, str]:
...
-    # Leave the symlinks in the build directory. This helps with reproducing.
-    symlink_dir = build_dir / "pywasmcross_symlinks"
-    shutil.rmtree(symlink_dir, ignore_errors=True)
-    symlink_dir.mkdir()
-    yield symlink_dir
+def _create_symlink_dir(build_dir: Path | None) -> Path:
+    if build_dir is None:
+        # If no build directory is provided, create a temporary directory
+
+        # NOTE: This is not ideal and should not be merged as-is. We need
```
hoodmane, line 291: Is there anything that wrong with this? Temporary directories get removed on reboot anyways. Also wouldn't the alternative be... using a context manager?

ryanking13:
Why not fix `get_build_env` to accept non-None path instead? I think all the callers now sets the build_dir properly.

Issue #169, opened 2025-04-02
Add an `uninstall-all`/`clean` sub-command for xbuildenvs


agriyakhetarpal:
One consequence of #148 is that the xbuildenvs are downloaded into the XDG cache or Library/Caches/ folder, which is not a convenient location for the end user to access. Multiple xbuildenvs can take up a bunch of space. For local development, the fact that we use a VCS-based versioning scheme means that I have a bunch of redundant xbuildenvs at any time on my machine per commit, which I need to go and either delete or uninstall one by one.

Thus, I would like to propose a command that removes them – as a wrapper over `pyodide xbuildenv uninstall`. We could call it:

- `pyodide xbuildenv uninstall-all` (similar to `pipx`, and closer in wording to `uninstall` which we already have); or
- `pyodide xbuildenv clean` (similar to `uv cache clean` or `pip cache purge`, as we're storing it in the caches directory)

I don't have a preference here, but I'm happy to hear opinions on either of the options.

hoodmane:
> the fact that we use a VCS-based versioning scheme

Could we also change this version less often? Or make a separate version that we only change when we change install_xbuildenv?

hoodmane:
I think I like `clean` better.

hoodmane:
We could call the folder `.pyodide-xbuildenv-1` and then rename it to `.pyodide-xbuildenv-2` if we change `CrossBuildEnvManager.install()`.

agriyakhetarpal:
Ah, okay, that sounds good to me. That seems like the approach `pytest` takes for its testing folders, which are named `pytest-of-username/test_run_number/`, with the difference that it stores them in temporary directories and then discards them.

PR #183, opened 2025-04-09
Fix regression in cross build package installation

ryanking13:
This fixes a bug that cross build packages were not correctly installed in the hostsitepackages. It was introduced in #167.

```
/tmp/build-env-gqqhqr9x/lib/python3.13/site-packages/numpy/_core/include/../lib/
libnpymath.a: archive member 'meson-generated_ieee754.c.o' is neither Wasm      
object file nor LLVM bitcode    
```

Think about this scenario:

1. We are building statsmodels (that depends on scipy) in-tree.
2. numpy is built first, as it is a dependency of scipy.
3. numpy is built. After the build, the cross build files of numpy are installed in hostsitepackages.
4. scipy is built, After the build, the cross build files of scipy are installed in hostsitepackages.
  4.a. Becuase of the `--upgrade` option, installing scipy will trigger numpy installation as well, overwriting the numpy package installed in step 3.
5. numpy in the hostsitepackages now contain native header files and libraries
6. statsmodels try to use numpy during the build step, but it finds native files, rasing an error.

### DIFF
--- a/pyodide_build/recipe/builder.py
+++ b/pyodide_build/recipe/builder.py
@@ -605,7 +605,29 @@ class RecipeBuilderPackage(RecipeBuilder):
                     [
                         "pip",
                         "install",
+                        # Upgrade the package in the host environment if there is a
+                        # older or newer version of the package already installed.
+                        # However, we don't want to replace the dependencies of the package,
+                        # since they may contain cross-build files as well.
+                        # For instance, numpy and scipy are both cross-build packages, but scipy depends on numpy.
+                        # Therefore, if we install numpy first and then scipy, installing scipy
+                        # will overwrite the cross build files in numpy.
                         "--upgrade",
+                        "--no-deps",
+                        "-t",
+                        str(host_site_packages),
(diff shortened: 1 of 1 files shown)


agriyakhetarpal (approved):
Thanks, @ryanking13! We were bound to come across this sooner or later. I wonder if we can add a test recipe in our integration tests related to this, as it wouldn't come up otherwise until we update the submodule.

SciPy is a heavy dependency to compile, so we could look at any other package which constructs a dependency graph similar to the one you've mentioned in the PR description. `galpy` looks like an ideal one – I see that it takes ~3 minutes to build (https://github.com/jobovy/galpy/actions/workflows/build_pyodide_wheels.yml).

ryanking13:
Thanks for the suggestion @agriyakhetarpal! Another option would be implementing `clean` command (https://github.com/pyodide/pyodide-build/issues/169), and telling people to simply clean the environment after upgrading the package and remove`--upgrade` option.

I'll merge this as-is and find some good package to test as a follow-up to unblock other PRs.

PR #181, opened 2025-04-08
Revert #144 for now

agriyakhetarpal:
This reverts commit 8c92fb9c7c486f3add2db65ab5f73aa4075efd98.

This venv was broken in https://github.com/pyodide/pyodide-build/pull/144, but the breakage wasn't noticed there, as the Pyodide venv tests are only executed with `[integration]` and the integration tests didn't run.

This PR temporarily reverts the change. The key issue as pointed out by Hood below is that we get a `OSError: [Errno 8] Exec format error:` – the approach fails because CPython's virtual environment mechanism requires the interpreter to be a symlink, as it uses realpath operations on the executable path to detect the venv's folder structure and locate the `pyvenv.cfg` file. Replacing the symlink with an exec-based script disrupts this detection process when executing `pip`, and we cannot cannot handle the nested shebang resolution required by the script chain.

The relevant error logs are as follows:

<details>
<summary>Details</summary>

```
____________________________ test_pip_install[pure] ____________________________

base_test_dir = PosixPath('/private/var/folders/qn/7t0vq3ts721cmgt0tgrtgzl80000gn/T/pytest-of-runner/pytest-0/pyodide_test_base')
packages = ['six']

    @pytest.mark.integration
    @pytest.mark.parametrize(
        "packages",
        [
            ["six"],  # pure
            ["numpy"],  # compiled
            ["six", "numpy"],  # mixed
        ],
        ids=["pure", "compiled", "both-pure-and-compiled"],
    )
    def test_pip_install(base_test_dir, packages):
        """Test that our monkeypatched pip can install packages into the venv"""
        venv_path = base_test_dir / "test_venv"
    
        venv.create_pyodide_venv(venv_path, [])
        venv_pip_path = venv_path / "bin" / "pip"
        assert venv_pip_path.exists(), "pip wasn't found in the virtual environment"
    
        for package in packages:
>           result = subprocess.run(
                [
                    str(venv_pip_path),
                    "install",
                    package,
                    "-v",
                    "--disable-pip-version-check",
                ],
                capture_output=True,
                text=True,
                check=False,
            )

pyodide_build/tests/test_venv.py:205: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[...]

PR #185, opened 2025-04-09
Make create_pip_script work with uv actually this time

hoodmane:
Weird hack to work around:
https://github.com/astral-sh/python-build-standalone/issues/380 

If we resolve the symlink all the way, the python-host interpreter works but won't install into our pyodide venv. If we don't resolve the symlink, sys.prefix is calculated incorrectly. To ensure that we get the right sys.prefix, we explicitly set it with the PYTHONHOME environment variable and then call the symlink.

### DIFF
--- a/pyodide_build/out_of_tree/venv.py
+++ b/pyodide_build/out_of_tree/venv.py
@@ -222,12 +222,14 @@ def create_pip_script(venv_bin):
     # Python in the shebang. Use whichever Python was used to invoke
     # pyodide venv.
     host_python_path = venv_bin / f"python{get_pyversion()}-host"
+    host_python_path_no_version = venv_bin / f"python-host"
     pip_path = venv_bin / "pip_patched"
     python_host_link = venv_bin / "python-host-link"
 
     # To support the "--clear" and "--no-clear" args, we need to remove
     # the existing symlinks before creating new ones.
     host_python_path.unlink(missing_ok=True)
+    host_python_path_no_version.unlink(missing_ok=True)
     python_host_link.unlink(missing_ok=True)
     for pip in venv_bin.glob("pip*"):
         if pip == pip_path:
@@ -253,7 +255,7 @@ def create_pip_script(venv_bin):
         )
     )
     host_python_path.chmod(0o777)
-    (venv_bin / "python-host").symlink_to(host_python_path)
+    host_python_path_no_version.symlink_to(host_python_path)
 
     pip_path.write_text(
         # Other than the shebang and the monkey patch, this is exactly what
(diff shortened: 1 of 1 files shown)


ryanking13 (approved):
Thanks for dealing with these issues.

agriyakhetarpal:
(still needs the integration test, BTW)

joerick:
I think the pip shebang issue on macOS can be fixed by executing that with `/usr/bin/env` i.e. `/usr/bin/env /path/to/python-host` - the reason it doesn't work as-is is that macOS expects script interpreters to be actual executables, not themselves scripts. But the `/usr/bin/env` program passes that through fine.

However, I don't know how to address the linux problem! My assumption is that pip is succeeding here, but not installing into the right venv?

hoodmane:
That does look like what happened. But I can try it on local dev and fix it manually. OTOH, I don't have a mac so your advice there is very much appreciated.

hoodmane:
Okay, I fixed at least a few problems.

hoodmane:
Okay, it's green. Gonna go ahead and merge this. Thanks again for your help @joerick!

agriyakhetarpal:
Thanks a lot, @hoodmane!

hoodmane:
And via manual testing I can confirm that it now works for me when `pyodide-build` is installed in a uv-managed venv.

agriyakhetarpal:
Sure, @hoodmane. Though, since we didn't include the previously incomplete fix in 0.30.0 (as we reverted it in #181 before releasing), should we create a 0.31 release instead?

hoodmane:
Not sure what our versioning rule is. But if it's semver-esque, this change itself is not a breaking change. So unless there is another breaking change on the main branch that you're worried about, I'd call it 0.30.1.

agriyakhetarpal:
Ah, okay, makes sense to me. Triggering a new release, then!

agriyakhetarpal:
Actually, I checked and we do have some new changes merged after 0.30.0 that are strictly features (#186 and #187). So, coupled with this change, it makes for two user-facing features and one bug fix. Given that we don't have branches set up to backport this change—and maybe we don't need them either, unlike how we do for the main Pyodide repo—can we go ahead with 0.31.0?

agriyakhetarpal:
Actually, I checked and we do have some new changes merged after 0.30.0 that are strictly features (#186 and #187). So, coupled with this change, it makes for two user-facing features and one bug fix. Given that we don't have branches set up to backport this change—and maybe we don't need them either, unlike how we do for the main Pyodide repo—could we go ahead with 0.31.0, if you're okay with it?

joerick:
I've tried this in pypa/cibuildwheel#2002 (installing from `main`) and it works! The only thing that doesn't quite work yet are package entrypoints like `pytest` - for those I get the same error. I'm guessing the approach used for `pip` here will work there too?

PR #191, opened 2025-04-12
Allow running integration tests with an "integration" label

agriyakhetarpal:
At the moment, we run the integration tests only when the last the message of the last pushed commit to a PR's branch contains the `[integration]` in its string, and sometimes we might forget to add it, meaning that the integration tests are either not run or that we need to push another commit solely to trigger them again (and sometimes this commit is empty – I frequently do it). This PR makes them run on a persistent basis – the introduced condition will allow us to run them without an explicit commit, and the tests will keep running until the label exists. Removing the label and pushing a subsequent commit will not run them again.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - main
   pull_request:
+    types: [opened, synchronize, reopened, labeled]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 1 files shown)


agriyakhetarpal:
As I didn't add the integration label to the PR, the tests were skipped. When I'll introduce it and close/reopen the PR, they should run.

agriyakhetarpal:
and now they do!

agriyakhetarpal:
I have to note that this configuration means that the collaborator will still need to add another commit after they add the label to trigger the pipelines again. We can improve this by running the `pull_request:` trigger on labeled PRs, too. Let me remove the label and try that out.

agriyakhetarpal:
I have removed the label, and the tests should no longer run when I close/reopen.

ryanking13 (approved):
Thanks, using label looks much better.

agriyakhetarpal:
Thanks for the review!

PR #195, opened 2025-04-15
Build recipes from a shared directory [DRAFT]

agriyakhetarpal:


agriyakhetarpal:
Ah, never mind – in its current state, it extracts another directory under `recipes/numpy_shared_build/numpy-tests-2.2.3/`, which evades the caching we need. I will need to revise it a bit.

ryanking13:
Thanks @agriyakhetarpal. Please ping me when it is ready.

I wonder if we can utilize `source.path` field for this, by the way. It is [used to building packages using the local files](https://github.com/pyodide/pyodide-build/blob/71a6d09651814439feb161537888078f7ff68c19/pyodide_build/recipe/builder.py#L278-L286) instead of downloading them from URL.

It was used for 1) testing purpose, 2) and when micropip was inside pyodide/pyodide, but I guess we can use it for sharing the build directories as well. For example, `numpy` packages downloads the archive from URL, and `numpy-tests` sets the path to the numpy's build directory.

Of course, this is a hack, and I am not sure it is good for the long-term solution, but I think is can be considered as an option.

PR #197, opened 2025-05-02
Fix virtual environment cli entrypoints

hoodmane:
This fixes a regression introduced in #185. `sys.executable` for `pytest` changed value from `python-host` to `python-host-link`.
When we install a cli entrypoint, we need to set the shebang appropriately to `venv/bin/python` not to `venv/bin/python-host-link`.

### DIFF
--- a/pyodide_build/tests/test_venv.py
+++ b/pyodide_build/tests/test_venv.py
@@ -1,6 +1,7 @@
 import os
 import shutil
 import subprocess
+from textwrap import dedent
 
 import pytest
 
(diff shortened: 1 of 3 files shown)


pyodide_build/tests/test_venv.py
```diff
@@ -256,3 +258,36 @@ def test_pip_downgrade(base_test_dir):
     assert result.returncode == 0, result.stderr
     assert result.stdout.startswith("pip 24.0")
     assert venv_pip_path.readlink() == venv_pip_path.with_name("pip_patched")
+
+
+def test_pytest_invoke(base_test_dir):
```
ryanking13, line 263:
```suggestion
@pytest.mark.integration
def test_pytest_invoke(base_test_dir):
```

I think this test needs an integration marker.

ryanking13 (approved):
Thanks. Could you add a changelog?

PR #199, opened 2025-05-08
Add 'pyodide config get interpreter'

hoodmane:
Returns the path in the xbuildenv to the interpreter.


agriyakhetarpal (approved):
Thanks, @hoodmane! Could you please add an entry in the key-value combos in the config tests, and a CHANGELOG entry?

hoodmane:
Doesn't look to me like there are specific tests for this.

agriyakhetarpal:
> Doesn't look to me like there are specific tests for this.

They are here:

https://github.com/pyodide/pyodide-build/blob/7126c5f1bd56ed7f40ce29cbcbe588267a42434c/pyodide_build/tests/test_config.py#L140-L177

We just need to add an entry for both.

PR #200, opened 2025-05-08
Fix handling for unknown config variable in `pyodide config get`

hoodmane:


pyodide_build/cli/config.py
```diff
@@ -47,7 +49,7 @@ def get_config(
     configs = _get_configs()
 
     if config_var not in configs:
-        typer.echo(f"Config variable {config_var} not found.")
-        typer.Exit(1)
+        print(f"Config variable {config_var} not found.", file=sys.stderr)
+        raise typer.Exit(1)
```
ryanking13, line 53: 👍

pyodide_build/cli/config.py
```diff
@@ -47,7 +49,7 @@ def get_config(
     configs = _get_configs()
 
     if config_var not in configs:
-        typer.echo(f"Config variable {config_var} not found.")
-        typer.Exit(1)
+        print(f"Config variable {config_var} not found.", file=sys.stderr)
```
ryanking13, line 52: What's the purpose of changing typo.echo to print? If the goal is to avoid using typer APIs we might change it everywhere.

hoodmane: The point is to print to stderr.

agriyakhetarpal: The docs at https://typer.tiangolo.com/tutorial/printing/?h=stderr#printing-to-standard-error say that we need to use `stderr=True`, but it has to be done at the `rich.console.Console()` level. Maybe we should set that somewhere?

ryanking13 (approved):
It looks like some tests need to be updated to check `result.stderr`. Otherwise looks good to me.

agriyakhetarpal (approved):
Thanks, @hoodmane! I opened #204 to talk about Typer-related issues, so I think this should be good for now.

Issue #203, opened 2025-05-16
HTTP 429 error when doing `pyodide xbuildenv search`


joerick:
Over in pypa/cibuildwheel#2002 I saw a test failure due to this exception:

```
  + pyodide xbuildenv search --json --all
  ----------------------------- Captured stderr call -----------------------------
  ╭───────────────────── Traceback (most recent call last) ──────────────────────╮
  │ /tmp/cibw-run-3ba5e_80/cp312-pyodide_wasm32/build/venv/lib/python3.12/site-p │
  │ ackages/pyodide_build/cli/xbuildenv.py:169 in _search                        │
  │                                                                              │
  │   166 │                                                                      │
  │   167 │   # TODO: cache the metadata file somewhere to avoid downloading it  │
  │   168 │   metadata_path = metadata_path or cross_build_env_metadata_url()    │
  │ ❱ 169 │   metadata = load_cross_build_env_metadata(metadata_path)            │
  │   170 │   local = local_versions()                                           │
  │   171 │                                                                      │
  │   172 │   if show_all:                                                       │
  │                                                                              │
  │ /tmp/cibw-run-3ba5e_80/cp312-pyodide_wasm32/build/venv/lib/python3.12/site-p │
  │ ackages/pyodide_build/xbuildenv_releases.py:239 in                           │
  │ load_cross_build_env_metadata                                                │
  │                                                                              │
  │   236 │   │                                                                  │
  │   237 │   │   with _suppress_urllib3_logging():                              │
  │   238 │   │   │   with requests.get(url_or_filename) as response:            │
  │ ❱ 239 │   │   │   │   response.raise_for_status()                            │
  │   240 │   │   │   data = response.json()                                     │
  │   241 │   │                                                                  │
  │   242 │   │   return CrossBuildEnvMetaSpec.model_validate(data)              │
  │                                                                              │
  │ /tmp/cibw-run-3ba5e_80/cp312-pyodide_wasm32/build/venv/lib/python3.12/site-p │
  │ ackages/requests/models.py:1024 in raise_for_status                          │
  │                                                                              │
  │   1021 │   │   │   )                                                         │
  │   1022 │   │                                                                 │
  │   1023 │   │   if http_error_msg:                                            │
  │ ❱ 1024 │   │   │   raise HTTPError(http_error_msg, response=self)            │
  │   1025 │                                                                     │
  │   1026 │   def close(self):                                                  │
[...]

PR #206, opened 2025-05-19
Update `DEFAULT_CROSS_BUILD_ENV_METADATA_URL` to JSON file from GitHub Pages site

agriyakhetarpal:
This PR is related to #203, pyodide/pyodide#5642, and pypa/cibuildwheel#2002; it updates the `DEFAULT_CROSS_BUILD_ENV_METADATA_URL` to use the JSON file from the GitHub Pages site set up just now, rather than the raw GitHub URL. This is being done in order to circumvent rate limits provisioned by the GitHub API. The next step after this, in a follow-up, would be to cache this even further, so that we don't access the URL repeatedly.

I would like to propose that we create a new 0.30.4 release after this PR. I can do it of course, but I'd like to receive reviews on this, i.e., if this URL seems fine to you, or if I should go back to the Pyodide repo and change it to something else. Thank you! :)

cc: @joerick

### DIFF
--- a/pyodide_build/xbuildenv_releases.py
+++ b/pyodide_build/xbuildenv_releases.py
@@ -6,7 +6,9 @@ from functools import cache
 from packaging.version import Version
 from pydantic import BaseModel, ConfigDict
 
-DEFAULT_CROSS_BUILD_ENV_METADATA_URL = "https://raw.githubusercontent.com/pyodide/pyodide/main/pyodide-cross-build-environments.json"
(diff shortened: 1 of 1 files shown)


ryanking13 (approved):
Thanks!

> I would like to propose that we create a new 0.30.4 release after this PR. I can do it of course, but I'd like to receive reviews on this, i.e., if this URL seems fine to you, or if I should go back to the Pyodide repo and change it to something else. Thank you! :)

Sounds good for my side.

Issue #88, opened 2025-01-23
`pyodide venv` should take an abi version / index argument


hoodmane:
We can see if we can imitate a subset of virtualenv's arguments. I'm not even sure how `pyodide venv` decides which index to download from since we stopped versioning pyodide-build and the pyodide runtime together.

ryanking13:
Maybe we can use the [Pyodide Anaconda Index](https://anaconda.org/pyodide/dashboard). It still does not support CORS headers but `pyodide venv` does not have CORS constraints (as far as I know?), and it use pip-resolver to download packages.

hoodmane:
Isn't that somewhat orthogonal to being able to select the abi?

ryanking13:
> Isn't that somewhat orthogonal to being able to select the abi?

We can put packages with different ABI (`pyodide_2024_0` / `pyodide_2025_0`) in this index together. pip will select the packages with matching ABI.

hoodmane:
Ah interesting. But what about if they want the package set specifically from 0.26.4? Maybe nobody wants that?

PR #207, opened 2025-05-20
Add test entries for `PYODIDE_INTERPRETER` and `PYODIDE_PACKAGE_INDEX`

agriyakhetarpal:
This PR should add a few test entries for #199.

cc: @hoodmane

### DIFF
--- a/pyodide_build/tests/test_config.py
+++ b/pyodide_build/tests/test_config.py
@@ -47,8 +47,6 @@ class TestConfigManager:
                                   skip_emscripten_version_check = "1"
                                   xbuildenv_path = "my_custom/xbuildenv_path"
                                   ignored_build_requirements = "cmake foo bar"
-                                  "interpreter" = "$(PYODIDE_INTERPRETER)"
-                                  "package_index" = "$(PYODIDE_PACKAGE_INDEX)"
                                   """)
 
         config_manager = ConfigManager()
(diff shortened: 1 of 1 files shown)


pyodide_build/tests/test_config.py
```diff
@@ -43,6 +47,8 @@ def test_load_config_from_file(self, tmp_path, reset_env_vars, reset_cache):
                                   skip_emscripten_version_check = "1"
                                   xbuildenv_path = "my_custom/xbuildenv_path"
                                   ignored_build_requirements = "cmake foo bar"
+                                  "interpreter" = "$(PYODIDE_INTERPRETER)"
+                                  "package_index" = "$(PYODIDE_PACKAGE_INDEX)"
```
ryanking13, line 51: I don't think we expect people to overwrite these configs.

agriyakhetarpal: Yes, this is with the in-tree case, I'll drop it. What should be the scenario with the out-of-tree case?

ryanking13: I don't think people would want to overwrite these variables in out-of-tree as well. The interpreter is coming from the xbuildenv, and the package_index is created by `pyodide venv`. I guess these variables are used mostly for getting the path for these values rather than overwriting it?

ryanking13 (approved):
LGTM

agriyakhetarpal:
I agree with you here, @ryanking13 and @hoodmane does too – I don't see anyone in the near future trying to override the package index without horribly breaking their ABI compatibility due to a runtime, and we don't have a nice way to deal with this until #88, so I would be fine with closing this. We already have tests that check the existence of the keys and vars, so this might be redundant. Thanks for the review!

PR #208, opened 2025-05-20
Stop seeding `wheel` for `pyodide venv`

agriyakhetarpal:
The `--wheel`/`--no-wheel` options for were first removed from `pypa/virtualenv` in pypa/virtualenv#2868, added back in pypa/virtualenv#2876, and both PRs were released in https://github.com/pypa/virtualenv/releases/tag/20.31.0. These were then reintroduced in pypa/virtualenv#2884 via https://github.com/pypa/virtualenv/releases/tag/20.31.2, but are present only when using `virtualenv` on Python 3.8 and earlier versions, and don't show up in the CLI `--help` on later versions altogether as there is no effect. As we support Python >=3.12, we should be fine with dropping this option completely. I've also added a minimum bound for `virtualenv`, and we should be fine with dropping the option as I don't think anyone was using it anyway.

I noticed this as some of the integration tests have been failing on `main` and weren't caught because we didn't need to run them on the last few PRs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.30.4] - 2025/05/20
 
+- Fixed compatibility with `virtualenv` 20.31 and later. The Pyodide virtual environment via `pyodide venv` no longer seeds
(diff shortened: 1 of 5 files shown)


agriyakhetarpal:
Merging this for now as it's a trivial bug fix and doesn't add any new code. Thank you!

PR #210, opened 2025-05-23
Don't raise an error if there are spaces in file paths supplied to `PIP_CONSTRAINT`

agriyakhetarpal:
## Description

In the file https://github.com/pyodide/pyodide-build/blob/7126c5f1bd56ed7f40ce29cbcbe588267a42434c/pyodide_build/build_env.py#L339 and in the `RecipeBuilder` class, we were checking if the `PIP_CONSTRAINT` supplied contains spaces and we were raising a `ValueError`, and this came up in https://github.com/pypa/cibuildwheel/pull/2002#discussion_r2092804165. The issue exists with `pip`: https://github.com/pypa/pip/issues/10114; however, there is a reasonable workaround – we can use a URI in order to **not** let a space in the file path be interpreted as a separate file.

cc: @joerick @hoodmane

### DIFF
--- a/pyodide_build/build_env.py
+++ b/pyodide_build/build_env.py
@@ -334,13 +334,16 @@ def _create_constraints_file() -> str:
     if not constraints:
         return ""
 
-    if len(constraints.split(maxsplit=1)) > 1:
-        raise ValueError(
-            "PIP_CONSTRAINT contains spaces so pip will misinterpret it. Make sure the path to pyodide has no spaces.\n"
-            "See https://github.com/pypa/pip/issues/13283"
-        )
-
+    # If a path to a file specified PIP_CONSTRAINT contains spaces, pip will misinterpret
+    # it as multiple files; see https://github.com/pypa/pip/issues/13283
+    # We work around this by converting the path to a URI.
     constraints_file = Path(constraints)
+    constraints = (
+        constraints_file.as_uri()
+        if " " in str(constraints_file)
+        else str(constraints_file)
+    )
+
     if not constraints_file.is_file():
         constraints_file.parent.mkdir(parents=True, exist_ok=True)
         constraints_file.write_text("")
--- a/pyodide_build/build_env.py
+++ b/pyodide_build/build_env.py
@@ -15,6 +15,7 @@ from packaging.tags import Tag, compatible_tags, cpython_tags
 
 from pyodide_build import __version__
 from pyodide_build.common import default_xbuildenv_path, search_pyproject_toml, to_bool
+from pyodide_build.logger import logger
 
 RUST_BUILD_PRELUDE = """
 rustup default ${RUST_TOOLCHAIN}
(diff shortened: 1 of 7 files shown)


agriyakhetarpal:
Would it be alright to release 0.30.5 with #209 and this PR?

pyodide_build/build_env.py
```diff
@@ -334,13 +334,16 @@ def _create_constraints_file() -> str:
...
-            "PIP_CONSTRAINT contains spaces so pip will misinterpret it. Make sure the path to pyodide has no spaces.\n"
-            "See https://github.com/pypa/pip/issues/13283"
-        )
-
+    # If a path to a file specified PIP_CONSTRAINT contains spaces, pip will misinterpret
+    # it as multiple files; see https://github.com/pypa/pip/issues/13283
+    # We work around this by converting the path to a URI.
     constraints_file = Path(constraints)
+    constraints = (
+        constraints_file.as_uri()
```
ryanking13, line 342: How about extracting this out to a function so we don't need to have a same code here and in builder.py?

ryanking13 (approved):
Thanks! I left a minor comment.

pyodide_build/build_env.py
agriyakhetarpal: Yes, makes sense to me. I'll put it in `common.py`.

agriyakhetarpal: Done, thanks!

ryanking13 (approved):

joerick:
I'm a little confused. If somebody outside pyodide_build is setting the environment variable PIP_CONSTRAINT to `/path to file/with spaces.txt`, then that's a misconfiguration - pip would fail to understand this.

It looks to me that pyodide-build will take the above and transform it into `file:///path%20to%20file/with%20spaces.txt`. That's potentially helpful(?), but also that means that users can't pass multiple files like `file1.txt file2.txt`, because pyodide-build would transform it into `file://file1.txt%20file2.txt`.

I'm not sure why pyodide-build needs to do any transformation to the value of PIP_CONSTRAINT. Is there a version of this where pyodide-build just leaves the value alone?

agriyakhetarpal:
Thanks for your thoughts, @joerick – this was also my source of confusion; I thought that we were just wanting to accept spaces with file paths to accommodate `cibuildwheel`, but on re-reading `combine_constraints()`, it looks like it is combining `cibuildwheel`'s constraints with the user's constraints, and the fact that `pyodide-build` is not allowing constraints with spaces is part of the problem.

I cannot think of a neat way to allow _multiple_ constraints files with `PIP_CONSTRAINT` and still allow a file with spaces to be interpreted correctly, so I'll go ahead and remove the workaround from here.

So, would I be right here if I say that the implementation should be:

1. If a user supplies a file via `PIP_CONSTRAINT` to `pyodide-build` and there are no spaces, we proceed in the normal fashion, and `cibuildwheel` can combine/extend it with its constraints if it wants to
2. If a user wants to supply multiple constraints files to `pyodide-build`, they use `PIP_CONSTRAINT="my_constraints1.txt my_constraints2.txt"` as usual and `pip` interprets them as multiple files
3. If a user wants to supply a constraint file to `pyodide-build` whose name/path contains spaces, _they_ are the ones to implement and use the URI workaround themselves, and we shouldn't operate on the `PIP_CONSTRAINT` file ourselves. `cibuildwheel` can then extend the `PIP_CONSTRAINT` it receives from `pyodide-build` correctly as it will have a `file://` in the string, and combine its own constraints without issues.

with the assumption that our intention with this PR is to undo https://github.com/pypa/cibuildwheel/blob/e92147f52a3f32818588e968f4be9315c8c35635/test/test_environment.py#L103-L108?

pyodide_build/build_env.py
```diff
@@ -334,10 +335,16 @@ def _create_constraints_file() -> str:
...
+    # warn them about it and ask to convert the path to a URI or to remove spaces in it.
+    if " " in constraints:
+        logger.info(
+            "The value of PIP_CONSTRAINT contains spaces, and pip will interpret it as "
+            "multiple files. If you are using a single constraints file and it has spaces "
+            "in its file path, please convert it to a URI instead. Please ignore this "
+            "message if you are using multiple constraints files."
         )
 
     constraints_file = Path(constraints)
```
joerick, line 350: I suppose you'd have to split and take the first entry if you're keeping the following code? `constraints` might contain two files, so `Path(constraints)` wouldn't refer to a file.

ryanking13:
Huh, I didn't know PIP_CONSTRAINT accepts multiple constraint files.

agriyakhetarpal:
This PR has conflicts, and I didn't get the time to come back to it after this. Looks like #291 and #292 handled `PIP_BUILD_CONSTRAINT`.

PR #211, opened 2025-06-11
Fixes temporary file issue on Windows

dom96:
On Windows we cannot reopen a temp file so we get permission denied errors in this code. This PR resolves this issue.

CC @hoodmane

### DIFF
--- a/pyodide_build/common.py
+++ b/pyodide_build/common.py
@@ -619,8 +619,8 @@ def download_and_unpack_archive(
     #         f"Failed to download cross-build environment from {url} (status code: {r.status_code})"
     #     )
 
-    with NamedTemporaryFile(suffix=".tar") as f:
-        f_path = Path(f.name)
+    with TemporaryDirectory() as temp_dir:
+        f_path = Path(temp_dir) / "temp.tar"
         f_path.write_bytes(data)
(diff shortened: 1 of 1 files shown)


hoodmane:
How about:
```
    with TemporaryDirectory() as temp_dir:
        f_path = Path(temp_dir) / "temp.tar"
        f_path.write_bytes(data)
        with warnings.catch_warnings():
            # Python 3.12-3.13 emits a DeprecationWarning when using shutil.unpack_archive without a filter,
            # but filter doesn't work well for zip files, so we suppress the warning until we find a better solution.
            # https://github.com/python/cpython/issues/112760
            warnings.simplefilter("ignore")
            shutil.unpack_archive(str(f_path), path)
```

hoodmane:
@ryanking13 the comment is a bit perplexing since we're unpacking a tar file here. I guess we could add an explicit filter and remove `catch_warnings()` now?

ryanking13:
> @ryanking13 the comment is a bit perplexing since we're unpacking a tar file here. I guess we could add an explicit filter and remove catch_warnings() now?

I think I had some issue with setting the filter, but I don't exactly remember. I am fine to remove it if anything works.

PR #216, opened 2025-07-01
NFC some perf changes for pywasmcross

### LINKED ISSUE #215: Optimise `pywasmcross`
It is not related to this PR, but I think we can do some micro optimization here:

1. Let's change the list to a set for fast check
2. Initialize the set before for loop

_Originally posted by @ryanking13 in https://github.com/pyodide/pyodide-build/pull/168#discussion_r2174912287_

agriyakhetarpal:
Closes #215

A lot of these are very minor changes and are not intended to change the cross-compilation behaviour in any way. I've added in-line comments about them below. The idea is that it's not a bottleneck anywhere, so even if we end up saving 2-5 seconds per compilation, it is a win for us.

### DIFF
--- a/pyodide_build/pywasmcross.py
+++ b/pyodide_build/pywasmcross.py
@@ -110,10 +110,7 @@ def replay_genargs_handle_dashl(arg: str, used_libs: set[str], abi: str) -> str 
     """
     assert arg.startswith("-l")
 
-    if arg == "-lffi":
-        return None
-
-    if arg == "-lgfortran":
+    if arg in ("-lffi", "-lgfortran"):
         return None
 
     # Some Emscripten libraries that use setjmp/longjmp.
--- a/pyodide_build/pywasmcross.py
+++ b/pyodide_build/pywasmcross.py
@@ -116,7 +116,7 @@ def replay_genargs_handle_dashl(arg: str, used_libs: set[str], abi: str) -> str 
     # Some Emscripten libraries that use setjmp/longjmp.
     # The Emscripten linker should automatically know to use these variants so
     # this shouldn't be necessary.
-    if abi > "2025" and arg in ["-lfreetype", "-lpng"]:
+    if abi > "2025" and arg in ("-lfreetype", "-lpng"):
         arg += "-legacysjlj"
 
     # WASM link doesn't like libraries being included twice
--- a/pyodide_build/pywasmcross.py
+++ b/pyodide_build/pywasmcross.py
@@ -8,10 +8,17 @@ with wrappers that adjusting include paths and flags as necessary for
 cross-compiling and then pass the command long to emscripten.
 """
 
+import functools
 import json
 import os
+import re
+import shutil
+import subprocess
 import sys
+import tempfile
+from collections.abc import Iterable, Iterator
 from pathlib import Path
+from typing import Literal, NamedTuple
 
 from __main__ import __file__ as INVOKED_PATH_STR
 
@@ -55,11 +62,6 @@ if IS_COMPILER_INVOCATION:
     )
     # restore __name__ so that relative imports work as we expect
     __name__ = PYWASMCROSS_ARGS.pop("orig__name__")
-
-
-import subprocess
-from collections.abc import Iterable, Iterator
-from typing import Literal, NamedTuple
 
 
 class CrossCompileArgs(NamedTuple):
@@ -349,8 +351,6 @@ def _calculate_object_exports_readobj_parse(output: str) -> list[str]:
 
 
 def calculate_object_exports_readobj(objects: list[str]) -> list[str] | None:
-    import shutil
-
     # This works for bootstrapped Emscripten via GitHub sources.
     # llvm-readobj might not be available this way with Homebrew
     # or conda-forge distributions of Emscripten.
@@ -458,8 +458,6 @@ def get_export_flags(
         export_list = exports
 
     prefixed_exports = ["_" + x for x in export_list]
-
-    import tempfile
 
     with tempfile.NamedTemporaryFile(mode="w", delete=False) as f:
         # Use a response file to avoid command line length limits
--- a/pyodide_build/pywasmcross.py
+++ b/pyodide_build/pywasmcross.py
@@ -40,6 +40,28 @@ SYMLINKS = {
     "install_name_tool",
     "otool",
 }
+
+EXCLUDED_LINKER_OPTS = {
+    "-Bsymbolic-functions",
+    # breaks emscripten see https://github.com/emscripten-core/emscripten/issues/14460
+    "--strip-all",
+    "-strip-all",
+    # wasm-ld does not recognize some link flags
+    "--sort-common",
+    "--as-needed",
+    # macOS-specific linker flags that wasm-ld doesn't understand
+    "-headerpad_max_install_names",
+    "-dead_strip_dylibs",
+}
+
+EXCLUDED_LINKER_PREFIXES = {
+    "--sysroot=",  # ignore unsupported --sysroot compile argument used in conda
+    "--version-script=",
+    "-R/",  # wasm-ld does not accept -R (runtime libraries)
+    "-R.",  # wasm-ld does not accept -R (runtime libraries)
+    "--exclude-libs=",
+}
+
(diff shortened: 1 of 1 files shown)


pyodide_build/pywasmcross.py
```diff
@@ -8,14 +8,21 @@
 cross-compiling and then pass the command long to emscripten.
 """
 
+import functools
 import json
 import os
+import re
+import shutil
+import subprocess
```
agriyakhetarpal, line 16: I moved all module imports to the top of the file, rather than placing them within functions. It should not amount to anything and is rather to make my linter happy. :)

pyodide_build/pywasmcross.py
```diff
@@ -8,14 +8,21 @@
...
 import sys
+import tempfile
+from collections.abc import Iterable, Iterator
 from pathlib import Path
+from typing import Literal, NamedTuple
 
 from __main__ import __file__ as INVOKED_PATH_STR
 
-INVOKED_PATH = Path(INVOKED_PATH_STR)
+SHAREDLIB_REGEX = re.compile(r"\.so(.\d+)*$")
```
agriyakhetarpal, line 25: I haven't profiled how expensive this regular expression is or how much it is called – perhaps a thousand or so times in total as it's called at link time? It would make sense to store it as a variable.

pyodide_build/pywasmcross.py
```diff
@@ -33,6 +40,60 @@
...
+}
+
+EXCLUDED_ARG_PREFIXES = {
+    "-J",  # fortran flag that clang does not support
+}
+
+SYS_PREFIX_INCLUDE = sys.prefix + "/include/python"
+SYS_BASE_PREFIX_INCLUDE = sys.base_prefix + "/include/python"
+
+INVOKED_PATH = Path(INVOKED_PATH_STR)
```
agriyakhetarpal, line 96: As above; they are stored as variables. It does hinder navigability, so please let me know and I can move them closer to the methods where they are used.

pyodide_build/pywasmcross.py
```diff
@@ -86,14 +142,7 @@ def is_link_cmd(line: list[str]) -> bool:
...
     """
-    import re
-
-    SHAREDLIB_REGEX = re.compile(r"\.so(.\d+)*$")
-    for arg in line:
-        if not arg.startswith("-") and SHAREDLIB_REGEX.search(arg):
-            return True
-
-    return False
+    return any(not arg.startswith("-") and SHAREDLIB_REGEX.search(arg) for arg in line)
```
agriyakhetarpal, line 145: Return earlier

pyodide_build/pywasmcross.py
```diff
@@ -115,16 +164,13 @@ def replay_genargs_handle_dashl(arg: str, used_libs: set[str], abi: str) -> str
     """
     assert arg.startswith("-l")
 
-    if arg == "-lffi":
-        return None
-
-    if arg == "-lgfortran":
+    if arg in ("-lffi", "-lgfortran"):
```
agriyakhetarpal, line 167: Check both at once

pyodide_build/pywasmcross.py
```diff
@@ -115,16 +164,13 @@ def replay_genargs_handle_dashl(arg: str, used_libs: set[str], abi: str) -> str
...
         return None
 
     # Some Emscripten libraries that use setjmp/longjmp.
     # The Emscripten linker should automatically know to use these variants so
     # this shouldn't be necessary.
-    if abi > "2025" and arg in ["-lfreetype", "-lpng"]:
+    if abi > "2025" and arg in ("-lfreetype", "-lpng"):
```
agriyakhetarpal, line 173: Immutable + take less memory + we don't need this list elsewhere in the file

pyodide_build/pywasmcross.py
```diff
@@ -153,16 +199,19 @@ def replay_genargs_handle_dashI(arg: str, target_install_dir: str) -> str | None
     """
     assert arg.startswith("-I")
 
+    include_path_str = arg[2:]
+
```
agriyakhetarpal, line 203: Stored as variable and calculated once

pyodide_build/pywasmcross.py
```diff
@@ -233,42 +261,22 @@ def replay_genargs_handle_argument(arg: str) -> str | None:
...
-        "-J",  # fortran flag that clang does not support
-    )):
+    if arg.startswith(EXCLUDED_ARG_PREFIXES):
         return None
 
-    # fmt: on
     return arg
 
 
+@functools.cache
```
agriyakhetarpal, line 273: As I understand, the CMake compiler flags are calculated just once as this is a fixed mapping, the symlinks directory isn't changing, and they won't change between different invocations either.

agriyakhetarpal (commented):
Hi @ryanking13, I hope these optimisations are in line with what you expected. Thanks!

hoodmane:
If we had time we might consider rewriting `pywasmcross.py` in Rust. Of course Emscripten also uses both Python and Node to compile so I'm not sure about how much of the poor compilation performance comes from this, from Emscripten using Python and Node, or from clang having a hard time compiling C++.

hoodmane:
> Each compiler call gcc, g++, ld, etc will run a fresh new Python interpreter. Basically, we are running python pywasmcross.py for each command. So quite a lot of optimizations in this PR will not work for our case.

As another crazy too much work idea, we could consider making `pywasmcross.py` into a process server and reusing it. Though we'd probably need a pool or it would cause trouble with parallelism.

pyodide_build/pywasmcross.py
```diff
@@ -8,16 +8,23 @@
...
 from pathlib import Path
+from typing import Literal, NamedTuple
 
 from __main__ import __file__ as INVOKED_PATH_STR
 
-INVOKED_PATH = Path(INVOKED_PATH_STR)
+SHAREDLIB_REGEX = re.compile(r"\.so(.\d+)*$")
 
-SYMLINKS = {
+SYMLINKS = (
```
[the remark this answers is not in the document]
agriyakhetarpal: Makes sense to me! I read up on this a bit and found that the "construction time" for sets is actually larger: https://stackoverflow.com/a/45019699, while the lookup time for them is less than that for tuples. I thought it wouldn't make much of a difference because all the sets we made were mainly small, but I guess minor differences will add up in the end. I'll switch back to an approach where I store sets as variables so that we don't need to construct/initialise them when searching for an item within them.

agriyakhetarpal (commented):
Thanks for all of your comments, @ryanking13! I pushed a few changes to hopefully address your review comments, and I've added some comments below about the new changes.

Ironically, my impression was that `pywasmcross.py` was being called in the way @hoodmane suggests, i.e., it's initialised once and an interpreter of sorts is kept alive, which has already processed the script into bytecode, so we don't need to initialise the interpreter again. I think it's an interesting idea to rewrite it in Rust. I would have suggested Cython due to its greater compatibility and similar speed to C, but yes, profiling should be done before we proceed. We could use a different source-to-source compiler, say, [Nuitka](https://github.com/Nuitka/Nuitka/), so that we could embed it as a fast and standalone application containing a deepfrozen Python interpreter embedded in it, but that would also suffer from usage concerns inside a Python project also being used as a library, besides its CLI. Having a subinterpreter that is kept alive until a project compiles might be useful, but I'm wary of considering parallelism.

pyodide_build/pywasmcross.py
```diff
@@ -86,13 +83,21 @@ def is_link_cmd(line: list[str]) -> bool:
...
             return True
 
+        dot_so_pos = arg.find(".so.")
+        if dot_so_pos != -1:
+            version = arg[dot_so_pos + 4 :]
+            # versioned shared libraries start with a digit, and consist of only digits and dots.
+            if version and version[0].isdigit():
+                if all(c.isdigit() or c == "." for c in version):
+                    return True
+
```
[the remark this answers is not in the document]
ryanking13: Thanks. I think the current implementation is okay. The filenames will not be very long, so even if there are some micro-optimizations we can try, they won't be very effective.

ryanking13 (approved):
Thanks! It looks good to me. Could you please run integration test to make sure there are no regression?

PR #223, opened 2025-08-10
Add helper function for recipe clean up

### LINKED ISSUE #218: Add `pyodide build-recipes clean` subcommand for cleaning up the downloaded files
## Description

`pyodide build-recipes` is a command to build Pyodide recipes, such as recipes in [pyodide/pyodide-recipes](https://github.com/pyodide/pyodide-recipes). Currently, it does not support a way clean up the files that are downloaded from the URL, and those generated during the build step. I think it would be good to have one.

- `pyodide build-recipes clean`: clean up all the temporary files in the build directory.
- `pyodide build-recipes clean <pkg> <tag:...>` clean up only the files in the specified directory
  - Should we consider dependencies? I guess not.
- It should not remove the `dist` directory, which is used to store the built wheels
  - Or we can probably add some flags, something like `--all`, `--include-dist`, to clean up those directories as well?

teddygood:
Closes #218

### Summary
- Adds `clean` subcommand to `pyodide build-recipes` for removing downloaded sources and intermediate build artifacts.
- Preserves `dist/` by default; can include with `--include-dist`.
- Supports package names, tags (`tag:<name>`), or all packages (`--all`).
- No dependency traversal.

### Changes
- Implement `clean` command with:
  - Default: remove `<pkg>/build` and `<pkg>/build.log`.
  - `--include-dist`: also remove `<pkg>/dist` and global install dir.
- Target resolution via existing loader.
- Added CLI tests (`test_clean.py`).

### Usage
```bash
# Clean all (excludes dist/)
pyodide build-recipes clean

# Clean selected targets
pyodide build-recipes clean numpy tag:core

# Include dist/
pyodide build-recipes clean --include-dist

# Custom directories
pyodide build-recipes clean \
  --recipe-dir=PATH_TO_RECIPES \
  --build-dir=PATH_TO_BUILD \
  --install-dir=PATH_TO_DIST \
  PKG1 PKG2

### CODE REVIEW

**Review by ryanking13 — COMMENTED** [review]
Thanks for working on this @teddygood. I left some comments.

Because the CLI design would need some discussion, I think we can separate this PR. How about

1) first implementing some helper functions to clean up the recipe build artifacts,
2) then connect a CLI entrypoints to those functions.

While working on (1), you can probably also refactor some of our code to make it reusable. For instance, locating the log file or build directories are already implemented in [recipe](https://github.com/pyodide/pyodide-build/tree/main/pyodide_build/recipe) package, and hardcoding those directories are quite easy to break.

File: pyodide_build/cli/build_recipes.py [hunk 1]
```diff
@@ -261,6 +267,215 @@ def build_recipes(
     build_recipes_impl(packages, args, log_dir_, install_options)
 
 
+# Register subcommands to the Typer app
+app.command("no-deps")(build_recipes_no_deps)
+app.command("build")(build_recipes)
+
+
+@app.command("clean")
+def clean(
```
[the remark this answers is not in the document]
teddygood: I’ll split the CLI from the cleanup logic. The CLI will only parse args/defaults and call small, testable helpers (e.g., resolve targets, remove build/log, optional dist, optional global dist). Unit tests will focus on the helpers.

teddygood:
Thanks for the feedback! To proceed, here’s a small plan — please let me know if this looks good:

- Keep this PR limited to helper functions + unit tests (no CLI changes).
  - Reuse existing recipe utilities (no hardcoded paths).
  - Default: preserve dist; add an option in helpers to include dist when needed.
  - No dependency traversal; don’t implicitly include the “always” tag.

- Drop the unrelated commit (9673975) and remove subcommand/callback changes so current behavior stays the same.

- In a follow-up PR, wire the helpers to a top-level CLI:
  - `pyodide clean` / `pyodide clean recipes [<pkg>|tag:<name>]`
  - `--include-dist` to opt in removing per-package and global install dist.

If this direction sounds good, I’ll push the updates for this PR accordingly. Thanks!

ryanking13:
Thanks for summarizing the plans.

> Reuse existing recipe utilities (no hardcoded paths).

Yes, sounds good. Actually, some paths might be hardcoded in recipe classes as well, but I would like to avoid adding more of hardcoded paths. If you think you can refactor some code to make it more reusable, feel free to go ahead and refactor it. But please do not mix refactoring and feature implementation, unless the change is very trivial. Try to do it step-by-step.

> Default: preserve dist; add an option in helpers to include dist when needed.

Yes. Note that there are two dist directories, one for global dist directory that stores all the wheels and another for storing local wheels per package.

> No dependency traversal; don’t implicitly include the “always” tag.

This part can be a matter of choice. We can change the decision at any time later. So when implementing the cleanup feature, this part should be easy to modify.

> In a follow-up PR, wire the helpers to a top-level CLI:

Yes, sounds good.

> --include-dist to opt in removing per-package and global install dist.

Maybe something like `--all` would be better. Users do not need to know about the name `dist`. Anyway, the name can be changed easily, so let's discuss the name in a separate PR.

pyodide_build/recipe/cleanup.py
```diff
@@ -0,0 +1,118 @@
...
+    if names_or_tags is None:
+        names_or_tags = ["*"]
+
+    recipes = loader.load_recipes(
+        recipe_dir, names_or_tags, load_always_tag=include_always_tag
+    )
+    return list(recipes.keys())
+
+
+def remove_package_build(recipe_dir: Path, build_dir_base: Path, package: str) -> bool:
```
ryanking13, line 31: All the `remove_...` functions are doing the similar thing with duplicate codes, so it would be nice to refactor the functions to be more general.

I think we can have two functions that:

1. Removes any file or directory given as a parameter
2. Locate a set of files to remove

Then, the function 1 would have a more simple signature, and can be used in a more general purposes.

If we have more files to delete in the future, we can update function 2 to include those files.

pyodide_build/recipe/cleanup.py
```diff
@@ -0,0 +1,118 @@
...
+    build_dir: Path | None,
+    install_dir: Path | None,
+    targets: Iterable[str] | None,
+    include_dist: bool = False,
+    include_always_tag: bool = False,
+) -> int:
+    """
+    Clean recipe build artifacts and optionally dist directories.
+
+    Returns the number of items removed.
```
ryanking13, line 94: I am not sure if this function should return the number of items removed. It doesn't seem very useful to me.

pyodide_build/recipe/cleanup.py
```diff
@@ -0,0 +1,118 @@
...
+        logger.info("Removing %s", str(install_dir))
+        import shutil
+
+        shutil.rmtree(install_dir, ignore_errors=True)
+        return True
+    return False
+
+
+def perform_recipe_cleanup(
+    *,
```
ryanking13, line 83: I think we can make recipe_dir and targets as positional parameters, instead of making everything keyword-only parameters.

pyodide_build/tests/test_cleanup_helpers.py
```diff
@@ -0,0 +1,75 @@
...
+        (install_dir / "dummy.whl").write_text("wheel", encoding="utf-8")
+    return pkg_root, build_dir, dist_dir
+
+
+def test_perform_cleanup_preserves_dist_by_default(tmp_path: Path):
+    for pkg in ("pkg_test_graph1", "pkg_test_graph3"):
+        _make_pkg_with_artifacts(pkg)
+
+    removed = perform_recipe_cleanup(
+        recipe_dir=RECIPE_DIR,
```
ryanking13, line 35: You don't need to use a real recipe directory for this. I think you can utilize `tmp_path`.

pyodide_build/recipe/cleanup.py
```diff
@@ -0,0 +1,118 @@
...
+    if install_dir and install_dir.exists():
+        logger.info("Removing %s", str(install_dir))
+        import shutil
+
+        shutil.rmtree(install_dir, ignore_errors=True)
+        return True
+    return False
+
+
+def perform_recipe_cleanup(
```
ryanking13, line 82: naming: How about simply `clean_recipes`?

pyodide_build/cli/build_recipes.py
```diff
@@ -261,6 +267,215 @@ def build_recipes(
     build_recipes_impl(packages, args, log_dir_, install_options)
 
 
+# Register subcommands to the Typer app
+app.command("no-deps")(build_recipes_no_deps)
+app.command("build")(build_recipes)
```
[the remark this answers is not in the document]
agriyakhetarpal: Yes, having the possibility to extend `pyodide clean` SGTM, too.

pyodide_build/tests/test_clean.py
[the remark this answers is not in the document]
teddygood: Thanks. I forgot to remove it. I’ll remove it in this PR.

pyodide_build/recipe/cleanup.py
```diff
@@ -0,0 +1,118 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
teddygood: Thanks. There was no issue with type annotations. I added it out of habit. I’ll remove it.

pyodide_build/recipe/cleanup.py
teddygood: Agreed. I will remove the return value and make the function void, since the count is not particularly useful and no callers rely on it.

pyodide_build/tests/test_cleanup_helpers.py
teddygood: I think to avoid using a real recipe directory we should use pytest `tmp_path` to create an isolated temporary recipe directory and generate minimal `meta.yaml` files so that the loader can discover packages Is this approach okay for these tests?

ryanking13: Yes, I think it is okay.

pyodide_build/recipe/cleanup.py
```diff
@@ -0,0 +1,100 @@
...
+            import shutil
+
+            if path.exists():
+                logger.info("Removing %s", str(path))
+                shutil.rmtree(path, ignore_errors=True)
+        elif path.is_file():
+            logger.info("Removing %s", str(path))
+            path.unlink(missing_ok=True)  # type: ignore[call-arg]
+        else:
+            # Path does not exist; nothing to do
```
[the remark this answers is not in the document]
teddygood: Thanks, that makes sense. I’ll add a debug log for this case so it’s easier to trace when nothing happens.

pyodide_build/tests/test_cleanup_helpers.py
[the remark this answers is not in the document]
teddygood: Sure, I’ll rename it.

pyodide_build/recipe/cleanup.py
```diff
@@ -0,0 +1,100 @@
...
+    Locate filesystem paths to remove for a single package.
+    """
+    paths: list[Path] = []
+    # Per-package build directory
+    paths.append(build_dir_base / package / "build")
+    # Per-package build log
+    paths.append(recipe_dir / package / "build.log")
+    # Per-package dist directory (optional)
+    if include_dist:
+        paths.append(recipe_dir / package / "dist")
```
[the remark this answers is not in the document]
teddygood: Sure, I’ll work on this refactor in a follow-up PR. Thanks for the suggestion.

PR #225, opened 2025-08-16
Allow overriding the URL for the Rust Emscripten sysroot

agriyakhetarpal:
This PR exposes `rust_emscripten_target_url` via `pyodide config`. While it is not strictly needed, I imagine that it could be useful for those wanting to override this quickly for OOT builds.


ryanking13 (approved):
Thanks!

PR #227, opened 2025-08-17
Update dependencies

ryanking13:
- pyodide-cli: removed redundant version identifier
- pyodide-lock: relaxed the version restriction
- auditwheel-emscripten: updated version and also relaxed version restriction


agriyakhetarpal (approved):
Thanks! Please feel free to release this PR if you'd like to. Would you be okay with unpinning `pyodide-lock` and `auditwheel-emscripten` entire sometime, or is the intention to always regard `cibuildwheel` as an application and not a library?

PR #239, opened 2025-08-26
Install new xbuildenv if existing is for the wrong Python version

### LINKED ISSUE #238: New versions of pyodide-build fail on Python 3.12
Repro (latest version: 0.30.7):

```
$ uv venv .venv-workers --python python3.12
$ uv pip install -p .venv-workers/bin/python pip pyodide-build
$ ./.venv-workers/bin/pyodide venv ./.venv-workers/pyodide-venv
...
ValueError: local Python version (3.12) does not match the Python version (3.13) used to create the Pyodide cross-build environment. Please switch back to the original Python version, or reinstall the 
xbuildenv, by running `pyodide xbuildenv uninstall` and then `pyodide xbuildenv install`
```

Works fine on 0.30.2.

hoodmane:
There is a confusing behavior where depending on global state, the same command will either succeed or fail. If we run the first time and there is no installed xbuildenv, we will install an appropriate one. However, if we have an existing xbuildenv for the wrong Python version, we fail.

Resolves #238.

### DIFF
--- a/pyodide_build/build_env.py
+++ b/pyodide_build/build_env.py
@@ -86,10 +86,10 @@ def _init_xbuild_env(
     context = redirect_stdout(StringIO()) if quiet else nullcontext()
     with context:
         manager = CrossBuildEnvManager(xbuildenv_path)
-        matches, _ = manager.check_version_marker()
+        matches, _ = manager.version_marker_matches()
         if not matches:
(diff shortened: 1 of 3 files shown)


hoodmane:
Yeah working on adding one.

PR #243, opened 2025-10-15
Add install-emscripten CLI command

bulenty584:
- Add new 'pyodide xbuildenv install-emscripten' command
  - Add clone_emscripten() method to clone emsdk repository
  - Add install_emscripten() method to install and activate Emscripten SDK
  - Add tests for new functionality
  
Message echoed after installation complete for user to  show the path to the emsdk_env.sh so that user can activate it manually in their shell.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -107,7 +107,8 @@ jobs:
             { name: test-src, installer: uv },
             { name: test-integration-marker, installer: pip }, # installer doesn't matter
           ]
-        os: [ubuntu-latest, macos-latest]
+        # os: [ubuntu-latest, macos-latest]  # FIXME: https://github.com/oracle/graal/issues/11855
+        os: [ubuntu-latest]
         pyodide-version: [stable]
         include:
         # Run no-isolation tests and Pyodide minimum version testing only


pyodide_build/xbuildenv.py
```diff
@@ -404,6 +404,117 @@ def uninstall_version(self, version: str | None) -> str:
...
+            cwd=emscripten_root,
+        )
+
+        # Activate the specified Emscripten version
+        subprocess.run(
+            [
+                "./emsdk",
+                "activate",
+                "--embedded",
+                "--build=Release",
```
hoodmane: What's `--embedded` and `--build=Release` do?

pyodide_build/tests/test_install_emscripten.py
```diff
@@ -0,0 +1,656 @@
...
+
+import subprocess
+from unittest.mock import MagicMock, call
+
+import pytest
+
+from pyodide_build.xbuildenv import CrossBuildEnvManager
+
+
+class TestInstallEmscripten:
```
hoodmane: This seems like a lot more tests than we need.

pyodide_build/tests/test_install_emscripten.py
```diff
@@ -0,0 +1,366 @@
...
```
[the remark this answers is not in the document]
bulenty584: Ah yeah, you are completely correct. Removed!

pyodide_build/cli/xbuildenv.py
```diff
@@ -205,3 +205,27 @@ def _search(
         print(MetadataView.to_json(views))
     else:
         print(MetadataView.to_table(views))
+
+
+@app.command("install-emscripten")
+def _install_emscripten(
+    version: str = typer.Option(
+        "latest", help="Emscripten SDK Version (default: latest)"
```
ryanking13: The default should be the one in the xbuildenv. Pyodide requires the same Emscripten version that is used to build the Pyodide runtime.

You can call `get_build_flag("PYODIDE_EMSCRIPTEN_VERSION")` to get the proper version.

pyodide_build/xbuildenv.py
```diff
@@ -404,6 +404,116 @@ def uninstall_version(self, version: str | None) -> str:
...
+            Path to the emsdk directory inside the xbuildenv.
+        """
+        if not self.symlink_dir.exists():
+            raise ValueError(
+                "No active xbuildenv. Run `pyodide xbuildenv install` first."
+            )
+
+        xbuild_root = self.symlink_dir.resolve()
+        emsdk_dir = xbuild_root / "emsdk"
+        patches_dir = emsdk_dir / "patches"
```
ryanking13: I think the patches should be under `self.pyodide_root / "emsdk" / "patches"`

pyodide_build/xbuildenv.py
```diff
@@ -404,6 +404,116 @@ def uninstall_version(self, version: str | None) -> str:
 
         return version
 
+    def clone_emscripten(self, emsdk_dir: Path | str | None = None) -> Path:
```
ryanking13: Let's make this function private, to avoid using it directly outside.

```suggestion
    def _clone_emscripten(self, emsdk_dir: Path | str | None = None) -> Path:
```

ryanking13 (commented):
Thanks for working on this. Could you also update the integration test?

Currently we are installing [Emscripten in CI](https://github.com/pyodide/pyodide-build/blob/bdd0b1422064de6d92264a08c599e85001a28567/.github/workflows/main.yml#L163-L167) for integration test, but I think we can replace this with this new command to test if this new command works correctly.

pyodide_build/tests/test_install_emscripten.py
```diff
@@ -0,0 +1,370 @@
...
+
+    # 4. Activate emsdk
+    assert calls[3] == call(
+        ["./emsdk", "activate", "--embedded", "--build=Release", "latest"],
+        cwd=emsdk_dir,
+        check=True,
+    )
+
+
+def test_install_emscripten_patch_application(tmp_path, monkeypatch):
```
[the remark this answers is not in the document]
bulenty584: Noted and changed!

pyodide_build/cli/xbuildenv.py
```diff
@@ -205,3 +205,28 @@ def _search(
         print(MetadataView.to_json(views))
     else:
         print(MetadataView.to_table(views))
+
+
+@app.command("install-emscripten")
+def _install_emscripten(
+    version: str = typer.Option(
+        get_build_flag("PYODIDE_EMSCRIPTEN_VERSION"),
+        help=f"Emscripten SDK Version (default: ${get_build_flag('PYODIDE_EMSCRIPTEN_VERSION')})",
```
ryanking13: Let's not call this inside the function's default value. Calling `get_build_flag` requires installing xbuildenv and it is a heavy operation.

You can set the default to None, and call `get_build_flag` inside the function explicitly when the version is not given.

.github/workflows/main.yml
```diff
@@ -161,10 +161,7 @@ jobs:
           key: ${{ env.EMSDK_CACHE_NUMBER }}-${{ env.EMSCRIPTEN_VERSION }}-${{ runner.os }}
 
       - name: Install Emscripten
-        uses: mymindstorm/setup-emsdk@6ab9eb1bda2574c4ddb79809fc9247783eaf9021 # v14
-        with:
-          version: ${{ env.EMSCRIPTEN_VERSION }}
-          actions-cache-folder: ${{env.EMSDK_CACHE_FOLDER}}
+        run: pyodide install-emscripten
```
ryanking13: This should be `pyodide xbuildenv install-emscripten`.

Also as the current implementation does not activate the emsdk after running this command. It should be called manually.

pyodide_build/cli/xbuildenv.py
bulenty584: Oh ok, I took it out and added the default check in the function.

.github/workflows/main.yml
bulenty584: Ah yes, totally my bad. Added both in now!

pyodide_build/cli/xbuildenv.py
```diff
@@ -205,3 +205,31 @@ def _search(
         print(MetadataView.to_json(views))
     else:
         print(MetadataView.to_table(views))
+
+
+@app.command("install-emscripten")
+def _install_emscripten(
+    version: str = typer.Option(
+        None,
+        help="Emscripten SDK Version (default: PYODIDE_EMSCRIPTEN_VERSION from xbuildenv)",
```
ryanking13: Variable name like `PYODIDE_EMSCRIPTEN_VERSION` is not user-friendly (actually, it is an implementation detail that user doens't know).

So maybe let's just replace the help help text to such as:

```
Emscripten version corresponding to the target Pyodide version
```

.github/workflows/main.yml
```diff
@@ -161,10 +161,12 @@ jobs:
           key: ${{ env.EMSDK_CACHE_NUMBER }}-${{ env.EMSCRIPTEN_VERSION }}-${{ runner.os }}
```
ryanking13: Let's remove this emsdk caching job too. We can add back caching when we find a good way to cache pyodide-build.

bulenty584:
Hey @ryanking13 and @hoodmane, any updates on this? Thanks!

ryanking13:
> Should we just run this command in the background?

Yes, that would work too, and it is probably better as Emscripten might change what env variables are set between versions.

Issue #246, opened 2025-10-20
Emscripten.cmake to also provide `CMAKE_MODULE_LINKER_FLAGS` / `CMAKE_MODULE_LINKER_FLAGS_INIT`


Andrej730:
Currently `SIDE_MODULE` flags are only provided to `SHARED` type of libraries:
https://github.com/pyodide/pyodide-build/blob/6aece56a1813e92d46e9a2faab2ac6f92ef5f695/pyodide_build/tools/cmake/Modules/Platform/Emscripten.cmake#L105-L107

The issue I've stumbled upon moving my Python extension to emscripten - it's using `MODULE` type of llibrary and therefore don't get any `SIDE_MODULE` flags. 

`MODULE` is probably the most popular type of libraries for Python extensions. 
E.g. `Python_add_library` is using it - https://cmake.org/cmake/help/latest/module/FindPython.html#commands as a default and also `swig_add_library` (https://cmake.org/cmake/help/latest/module/UseSWIG.html#command:swig_add_library).

So possibly `SIDE_MODULE_LDFLAGS` flags can also be added for `CMAKE_MODULE_LINKER_FLAGS_INIT` to make transition to pyodide more seamless.

PS Though in my particular case it wouldn't help either way - I also have duplicated linked libraries conflicting with `SIDE_MODULE=1` and end up setting `SIDE_MODULE=2` with `target_link_options` as a workaround.

Andrej730:
A heads-up:
Migrating to 0.30.8 I've stumbled upon issue that `target_link_options(target PRIVATE "SHELL:-s SIDE_MODULE=2)` had no effect when cmake was using Makefile generator. 
On makefiles cmake seems to prepend those options, instead of appending, So `-s SIDE_MODULE=1` coming from `SIDE_MODULE_LDFLAGS` is overriding it. The odd thing that for Ninja it is appending options and everything worked as expected.


A workaround:
```cmake
    if(CMAKE_GENERATOR MATCHES "Makefile")
        # Accomodate bug in cmake makefile generator:
        # https://gitlab.kitware.com/cmake/cmake/-/issues/27326
        string(
            REPLACE "SIDE_MODULE=1" "SIDE_MODULE=2"
            CMAKE_MODULE_LINKER_FLAGS
            "${CMAKE_MODULE_LINKER_FLAGS}"
        )
    endif()
```

Reported it to cmake - https://gitlab.kitware.com/cmake/cmake/-/issues/27326

[...]

PR #249, opened 2025-10-26
Update cmake toolchain file to use `_INIT` variable for cflags and cxxflags as well

ryanking13:
Follow up of #246


Andrej730:
Any particular packages that break with this? Wanted to try to look into it.

PR #253, opened 2025-11-03
Feat/auto activate emsdk

bulenty584:
As discussed with @ryanking13, this pr automatically activates emscripten for users (e.g. running source emsdk.sh) and removes the need for them to do it manually. Changes are as follows:

    Added activation and error handling in xbuildenv.py
    Removed erroneous output after installtion in cli/xbuildenv.py

Previous PR was closed due to a branch mistake

### DIFF
--- a/pyodide_build/tests/test_cli_install_emscripten.py
+++ b/pyodide_build/tests/test_cli_install_emscripten.py
@@ -60,8 +60,6 @@ def test_install_emscripten_default_version(tmp_path, monkeypatch):
     assert result.exit_code == 0, result.stdout
     assert "Installing emsdk..." in result.stdout, result.stdout
     assert "Installing emsdk complete." in result.stdout, result.stdout
-    assert "Use `source" in result.stdout, result.stdout
-    assert "emsdk_env.sh` to set up the environment." in result.stdout, result.stdout
     assert called["version"] == "3.1.46"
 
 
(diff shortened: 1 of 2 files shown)


ryanking13:
Thanks for working on this @bulenty584! I have a few comments.

1. Calling `source` inside a subprocess does not affect the local shell. If you see the [`emsdk_env.sh`](https://github.com/emscripten-core/emsdk/blob/a0af5d7f952e6fbe5ffac1e05805bf07f450fabb/emsdk_env.sh#L72) you can find that it uses the `eval` trick to set the env variable for the local shell.

2. I am okay with running `emsdk_env.sh` after running `install-emscripten` command, but since it is volatile after the user closes the shell and reopens it, it would be nice to make a separate command for it, such as `pyodide xbuildenv activate-emscripten`.

3. Actually, what I really prefer as a long term solution is to automatically adding `emcc` to the path when a user runs `pyodide build` or `pyodide build-recipes`  without requiring the user to activate emcc manually. Of course, we can have both options (activating emcc explicitly or injecting path implicitly). But, the problem in (1) can be solved by storing the env variables as an internal variable in Python and passing them every time we call `subprocess`.

But you don't need to handle (3) in this PR. I am happy to accept this PR with (1) and (2) solved.

ryanking13:
> sorry I haven't been very active as I have had some midterms.

No worries :) We are all volunteers, and each person should work according to their own schedule.

> My understanding is the user must add relevant Emscripten env vars and adjust their path to point to the new emsdk but I am confused as to how adding emcc to their path achieves the same goal or functionality?

What `source emsdk_env.sh` does is adding some env vars to the shell. We are also modifying the env vars that we pass to the subprocess call [here](https://github.com/pyodide/pyodide-build/blob/5c01c9e0fa1a42a28207d9abb03144e4c651d819/pyodide_build/pypabuild.py#L344-L350). So if we update the env vars inside the Python code there, it basically has the same effect.

PR #261, opened 2025-11-28
Use python.bat for Pyodide python interpreter in Windows

ryanking13:
This will not properly work yet, but we didn't support Windows anyways. This is a preparation for pyodide-build side to make `pyodide venv` work. Rest of the work needs to be done (hopefully), in Pyodide.


agriyakhetarpal (commented):
Thanks, @ryanking13! Should this be `python.exe` instead of `.bat`? Windows venvs have a `python.exe` and a windowless `pythonw.exe`. I just wanted to verify this before we merge.

Also, could we add a test job in the matrix for this? We can skip all tests on Windows for now via the `pytest_collection_modifyitems` hook, and run only this test by explicitly marking it to run on Windows.

Issue #263, opened 2025-12-02
Adding windows support for `pyodide venv`


ryanking13:
This issue is to track the state of the Windows support for `pyodide venv`. Currently, it works only in Unix compatible environment.

- [x] Add Windows batchfile wrapper for Pyodide python CLI (https://github.com/pyodide/pyodide/pull/6012)
- [x] Replace all linux-specific path to be universal (#264)
- [x] Remove `make` dependency (https://github.com/pyodide/pyodide-build/pull/266)
- [x] Handle `EBADF: bad file descriptor, fsync` error in windows (https://github.com/pyodide/pyodide/pull/6018)
- [x] (Optional) Refactor the `venv.py` to be more extensible

PR #266, opened 2025-12-03
Add fallback method for parsing Makefile.envs

ryanking13:
Related: #263

Adds a fallback method for parsing the makefile. This seems to be enough for reading variables we use in `pyodide venv`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,7 +25,7 @@ jobs:
       fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           # include tags so that hatch-vcs can infer the version
           fetch-depth: 0
@@ -39,7 +39,7 @@ jobs:
           python-version: "3.12"  # TODO: update to 3.13
 
       - name: Set up Node.js
-        uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6.0.0
+        uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6.1.0
         with:
           node-version: "20"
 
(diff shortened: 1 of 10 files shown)


pyodide_build/config.py
```diff
@@ -122,25 +122,81 @@ def _get_make_environment_vars(self) -> Mapping[str, str]:
...
+        env = os.environ | {"PYODIDE_ROOT": str(self.pyodide_root)}
+        makefile_path = self.pyodide_root / "Makefile.envs"
+
+        # Try using make first if available
+        try:
+            result = run_command(
+                ["make", "-f", str(makefile_path), ".output_vars"],
+                env=env,
+                err_msg="ERROR: Failed to load environment variables from Makefile.envs",
+            )
```
hoodmane: This try/catch is broader than it needs to be. How about:
```
    result = None
    try:
        result = run_command(...)
    except CommandFailedException: # Not sure what we raise
        pass
    if result is None:
        logger.debug(
            "make command not available, using fallback Makefile.envs parser"
        )
        return _parse_makefile_envs(env=env, makefile_path=makefile_path)
```

pyodide_build/config.py
```diff
@@ -122,25 +122,81 @@ def _get_make_environment_vars(self) -> Mapping[str, str]:
...
+        for line in f:
+            line = line.strip()
+
+            # Skip comments and empty lines
+            if not line or line.startswith("#"):
+                continue
+
+            # Only process export statements
+            if line.startswith("export "):
+                line = line[7:].strip()  # Remove "export "
```
hoodmane:
```suggestion
                line = line.removeprefix("export ").strip()  # Remove "export "
```

pyodide_build/tests/test_config.py
```diff
@@ -216,3 +217,112 @@ def test_cli_config_subset():
         assert value in BUILD_VAR_TO_KEY, (
             f"All cli config values should be in build var to key mapping, but {value} is not"
         )
+
+
+class TestParseMakefileEnvs:
```
hoodmane: How about a test with an assignment like `A=2` that doesn't start with export? Also, `export A=2 # this is a comment` isn't handled though it seems to set `A` to `"2 "`.

pyodide_build/config.py
hoodmane: Or what about checking if the platform is windows and avoiding using a try/catch at all?

ryanking13: Makes sense.

> Or what about checking if the platform is windows and avoiding using a try/catch at all?

Changed to this way.

PR #269, opened 2025-12-10
Remove `-Z link-native-libraries=no` default

hoodmane:
We haven't wanted this for a long time

### DIFF
--- a/integration_tests/recipes/orjson/meta.yaml
+++ b/integration_tests/recipes/orjson/meta.yaml
@@ -9,8 +9,6 @@ source:
 requirements:
   executable:
     - rustup
-build:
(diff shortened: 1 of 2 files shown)


hoodmane:
We're overriding this in Makefile.envs:
https://github.com/pyodide/pyodide/blob/main/Makefile.envs#L15

ryanking13:
> We're overriding this in Makefile.envs:
https://github.com/pyodide/pyodide/blob/main/Makefile.envs#L15

Yes, but the config in `pyproject.toml` dominates the one in the Makefile.envs. So we can test set custom rustflags overwriting the default one.

PR #272, opened 2025-12-15
Use a global pip configuration inside xbuildenv

### LINKED ISSUE #271: Pyodide xbuildenv aren't pre-configured for use with commands other than `install`
If you create a Pyodide xbuildenv, `pip install` will "just work" in that environment, downloading binary wheels as required, and raising an error if a binary emscripten/pyodide wheel isn't available.

However, `pip wheel`, `pip download` or `pip lock` won't work; those commands will attempt to download a source tarball for a candidate package.

This is because the `pip.conf` that is generated when the xbuildenv is created explicitly configures the `install` command, rather than setting a global configuration. As a result, pip won't consider pyodide wheels as candidates for download for commands other that `pip install`.

freakboy3742:
Modifies the pip configuration generated by `xbuildenv` to use a global configuration, rather than a specific configuration of the `install` command. 

This means that `pip wheel numpy`, `pip download numpy` and `pip lock numpy` will download a binary wheel, rather than downloading a source wheel and trying to compile it.

Fixes #271.

### DIFF
--- a/pyodide_build/out_of_tree/venv.py
+++ b/pyodide_build/out_of_tree/venv.py
@@ -94,7 +94,7 @@ def create_pip_conf(venv_root: Path) -> None:
     (venv_root / "pip.conf").write_text(
         dedent(
             f"""
-            [install]
+            [global]
             only-binary=:all:
             {repo}
             """


ryanking13 (approved):
Makes sense. Thanks!

PR #275, opened 2025-12-16
Use nightly xbuildenv for integration test

ryanking13:
As title. I would like to run integration test with nightly xbuildenv for testing Windows venv.


agriyakhetarpal:
Hi @ryanking13, this one would be good to add. I've added an entry for the 314.0.x stable-debug xbuildenvs in #407, so we can use the new(er) `pyodide xbuildenv install --nightly` commands for this, instead of pinning the xbuildenv in `pyproject.toml`.

I merged `main` into this branch to rerun CI, as the logs have expired 😅

ryanking13:
yeah, thanks @agriyakhetarpal! I don't exactly remember why this failed before, but it shouldn't be very hard to add. I'll close this for now so that you can open your own PR if you are interested.

PR #282, opened 2026-01-06
Add option to clean build artifacts after recipe build

### LINKED ISSUE #279: Feature request: clean while building recipes
I am running out of space in CI when building recipes. I already have a post-build cleaning step, `ls -d packages/*/build/* | xargs rm -r` in the recipe dir. Maybe this could automatically be done for each recipe once it has been built?

I imagine having a `--clean` flag that, when enabled, removes the build directory after a successful build.

ryanking13:
Resolve #279

Reuses clean up functions implmented in #254 (cc: @teddygood)

- [x] changelog

### DIFF
--- a/pyodide_build/tests/test_cli.py
+++ b/pyodide_build/tests/test_cli.py
@@ -152,6 +152,55 @@ def test_build_recipe_clean(tmp_path, dummy_xbuildenv, mock_emscripten):
         assert not build_dir.exists(), f"Build directory should be cleaned: {build_dir}"
 
 
+def test_build_recipe_clean_custom_build_dir(
+    tmp_path, dummy_xbuildenv, mock_emscripten
+):
+    """Test that --clean with custom --build-dir cleans package build directory correctly."""
+    output_dir = tmp_path / "dist"
+    custom_build_dir = tmp_path / "custom_build_dir"
+    custom_build_dir.mkdir()
+
+    pkgs_to_build = ["pkg_test_graph1", "pkg_test_graph3"]
+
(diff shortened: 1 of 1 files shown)


agriyakhetarpal (approved):
Hi @ryanking13, thanks for this PR! I like the overall direction here. However, during testing, I believe this feature has revealed another related bug, though it is unrelated to your changes in this PR.

Currently, the `pyodide clean recipes` command has a `--build-dir` argument, which, IIUC, is meant to be supplied a top-level directory where per-package subdirectories are created. This is not related to your work in this PR, of course, but I think we should be consistent in both places: right now, say, we build a package using `pyodide build-recipes pydoc_data --build-dir custom_build_dir --clean`. This will build `pydoc_data` and, when done, clean the `pydoc_data/build` folder within the recipe directory.

However, if I were to not pass `--clean` and instead want to clean the builds later on, `pyodide clean recipes pydoc_data --build-dir custom_build_dir` deletes the entire `custom_build_dir`, instead of `custom_build_dir/pydoc_data/build/`. This means that if I were to have some other packages in `custom_build_dir` that I may have built, their artifacts are also cleaned up inadvertently. That is, `pyodide clean recipes <package> --build-dir <dir>` does not respect the package choice, in comparison to `pyodide build-recipes <package> --clean` here, which does. This is a bug with the implementation in #254, not here, so I can take that up in a separate PR.

So, regardless, on a similar note, could you please add one more test that runs `pyodide build-recipes --clean` with a custom build directory and verifies that the package build directory inside it is cleaned correctly?

PR #302, opened 2026-02-25
Install cross-build packages only when used

### LINKED ISSUE #297: Install cross-build packages only when they are actually used
Probably we should change the timing that installs the cross build package, so that we only install them when we actually use them.

_Originally posted by @ryanking13 in https://github.com/pyodide/pyodide-build/issues/296#issuecomment-3931195974_

bulenty584:
Fixes #297 

1. Install cross-build packages lazily: they are now installed only when first needed during build flow.
2. Added marker-based idempotency (.cross-build-packages-installed) to avoid repeated installs.
3. Updated pypabuild to trigger on-demand install when unisolated packages are missing.
4. Added tests for lazy trigger and idempotent behavior ([test_pypabuild.py](https://github.com/pyodide/pyodide-build/blob/main/pyodide_build/tests/test_pypabuild.py), [test_xbuildenv.py](https://github.com/pyodide/pyodide-build/blob/main/pyodide_build/tests/test_xbuildenv.py)).

### DIFF
--- a/pyodide_build/xbuildenv.py
+++ b/pyodide_build/xbuildenv.py
@@ -92,21 +92,6 @@ class CrossBuildEnvManager:
         """
         return version_path / CROSS_BUILD_PACKAGES_MARKER_FILE
 
-    def _current_version_path(self) -> Path:
-        """
-        Return the currently selected concrete xbuildenv version directory.
-
-        Raises
-        ------
-        ValueError
-            If no active xbuildenv is selected.
-        """
-        if not self.symlink_dir.exists():
-            raise ValueError(
-                "No active xbuildenv. Run `pyodide xbuildenv install` first."
-            )
-        return self.symlink_dir.resolve()
-
     def ensure_cross_build_packages_installed(self) -> None:
         """
         Install cross-build packages for the active xbuildenv only when needed.


pyodide_build/pypabuild.py
```diff
@@ -106,6 +106,26 @@ def symlink_unisolated_packages(env: DefaultIsolatedEnv) -> None:
...
+
+    needs_cross_build_install = any(
+        not any(
+            chain(
+                host_site_packages.glob(f"{name}*"),
+                host_site_packages.glob(f"_{name}*"),
+            )
+        )
+        for name in unisolated_packages
+    )
```
ryanking13: This will not work because the host_site_packages will be empty if we don't install the cross build packages in the first place.

I think the more robust way to check if we need to install them is by checking if the cross-build packages are included in the build requirements.

So I think we can

1) update the function signature to include `reqs`.
2) check if any unisolated packages are included in the `reqs`.
3) If so (and if we are in xbuildenv), install the packages to host_site_packages
4) do the rest of the process

pyodide_build/pypabuild.py
```diff
@@ -106,6 +109,20 @@ def symlink_unisolated_packages(env: DefaultIsolatedEnv) -> None:
...
+    required = {Requirement(req).name.lower() for req in (reqs or set())}
+
+    needs_cross_build_install = bool(unisolated_packages & required)
+
+    if in_xbuildenv() and needs_cross_build_install:
+        from pyodide_build.common import default_xbuildenv_path
+        from pyodide_build.xbuildenv import CrossBuildEnvManager
+
+        CrossBuildEnvManager(
+            default_xbuildenv_path()
```
ryanking13: The user might have instantiated the xbuildenv from non-default path, so we should not use `default_xbuildenv_path` here. Instead, I think we should locate the current root directory of xbuildenv

In your previous PR, we calculated it from pyodide-root, so I think we can use the same thing here. Maybe we can create a helper method (e.g. `get_current_xbuildenv_manager`) does this to avoid duplicated code.

https://github.com/pyodide/pyodide-build/blob/2dd2157f2e6728435909c6847e6bd97511c22491/pyodide_build/build_env.py#L374-L386

pyodide_build/pypabuild.py
```diff
@@ -106,6 +109,17 @@ def symlink_unisolated_packages(env: DefaultIsolatedEnv) -> None:
...
     shutil.copy(sysconfigdata_path, env_site_packages)
     host_site_packages = Path(get_hostsitepackages())
+    unisolated_packages = set(get_unisolated_packages())
+
+    required = {Requirement(req).name.lower() for req in (reqs or set())}
+
+    needs_cross_build_install = bool(unisolated_packages & required)
+
+    if in_xbuildenv() and needs_cross_build_install:
+        from pyodide_build.build_env import get_current_xbuildenv_manager
```
ryanking13: Let's just import this in the top level. Unless there is a concern of cyclic imports, there isn't much reason to dynamically import this.

pyodide_build/pypabuild.py
```diff
@@ -189,7 +203,22 @@ def _build_in_isolated_env(
...
+            req
+            for req in build_reqs
+            if Requirement(req).name.lower() not in base_req_names
+        }
+
+        # We run install_reqs here if there are new packages to install
+        # No need to install packages unless we have to install cross-build packages that are used
+        if new_build_reqs:
+            symlink_unisolated_packages(env, new_build_reqs)
+            install_reqs(build_env, env, new_build_reqs)
```
ryanking13: No, we don't need to do this. All the cross-build package should be already discovered and installed in the previous `symlink_unisolated_packages`. If it hasn't that should be a user fault (not using pyproject.toml, etc). Let's revert this part and simply the logic.

CHANGELOG.md
```diff
@@ -7,7 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-## [0.33.0] - 2026/02/26
+### Changed
+
+- Cross-build package installation is now lazy by default. `pyodide xbuildenv install` can skip eager installation, and cross-build packages are installed on first build-time use.
+- Added marker-based idempotency for cross-build package installation, so repeated checks do not reinstall packages once they are already present.
+
+### Added
+
+- Added internal marker file `.cross-build-packages-installed` and `CrossBuildEnvManager.ensure_cross_build_packages_installed()` to support on-demand installation.
+- Added tests covering lazy installation trigger path and idempotent installation behavior.
```
ryanking13: We only add user facing changes to the changelog. I think most of these are redundant. Maybe let's simply put

- `pyodide xbuildenv install` can skip eager installation, and cross-build packages are installed on first build-time use.

pyodide_build/tests/test_pypabuild.py
```diff
@@ -97,3 +97,169 @@ def test_get_build_env(tmp_path, dummy_xbuildenv):
...
+    )
+
+    class DummyEnv:
+        path = str(tmp_path / "venv")
+
+    pypabuild.symlink_unisolated_packages(DummyEnv(), reqs={"wheel"})
+    assert called["count"] == 0
+
+
+def test_build_in_isolated_env_skips_second_pass_when_no_new_dynamic_reqs(
```
ryanking13: We don't need to handle dynamic reqs (as I mentioned in other comment). Let's remove this.

PR #319, opened 2026-03-30
Update platform name to `pyemscripten`

ryanking13:
- [x] Changelog


ryanking13:
@agriyakhetarpal 

Oh yeah. Hmm, I am +1 for just making a release. I don't think that the platform name will be changed again. If that happens, let's just yank that version and release a new version.

PR #322, opened 2026-04-03
Support `pyodide xbuildenv install-emscripten` on Windows

Andrej730:
Hi! Added support for running `pyodide xbuildenv install-emscripten` command on Windows.


agriyakhetarpal (commented):
Neat! We don't have CI for Windows running here, so would you be keen to add it to the matrix and check if everything is good?

The Windows machine I have is pretty old, so all I can do is trust what the CI says.

ryanking13:
Thanks! I will review this more carefully when I have time, but overall it looks okay.

One question: which compilers are you using in Windows to build the Python package for Pyodide? We only proxy `gcc` not the Visual Studio toolchains, so I was wondering if you are using mingw, or something else.

ryanking13:
(Closed by mistake)

Andrej730:
Hi! Any changes needed?

agriyakhetarpal:
@Andrej730 hi! I am +1 on merging this, as the code looks great, but I still think we can add a small Windows CI job (in the existing workflow files). No need to run the full test suite or declare Windows compatibility/support at this point; as we discussed that's a big pile of work :-) It would be sufficient to just run `pyodide xbuildenv install-emscripten` and, in a subsequent step, check that `emcc` is available and can compile basic C code. Otherwise, we won't know whether this works reliably and won't break in the future.

Andrej730:
Sure, added a simple integration test - for `xbuildenv` it runs `install`, `install-emscripten`, compiles a simple program with `emcc` and runs it.

pyodide_build/xbuildenv.py
```diff
@@ -540,20 +540,22 @@ def install_emscripten(
...
         try:
-            subprocess.run(
-                f"cat {patches_dir}/*.patch | patch -p1 --verbose",
-                check=True,
-                shell=True,
-                cwd=emscripten_root,
-            )
+            for patch_file in patches_dir.glob("*.patch"):
+                subprocess.run(
+                    ["git", "apply", "--verbose", str(patch_file)],
```
ryanking13, line 560: Nice. Maybe we could apply this to other places we use `patch` so make our system more OS-independent

(no action required for this PR)

.github/workflows/main.yml
```diff
@@ -64,6 +64,52 @@ jobs:
           if-no-files-found: error
           include-hidden-files: true
 
+  windows-integration-test:
```
ryanking13, line 67: Could you add integration test check so that this test only run under integration test environment?

https://github.com/pyodide/pyodide-build/blob/93f4df986849bb6798a9b9f619a4cd4b25e49fa4/.github/workflows/main.yml#L125

ryanking13 (commented):
Sorry for the late review. Yes, I am also okay with merging this. I left a minor comment but otherwise looks good to me.

.github/workflows/main.yml
Andrej730: Added in ea611f0

.github/workflows/main.yml
```diff
@@ -64,6 +64,54 @@ jobs:
...
+
+      - name: Setup Python
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: "3.12"
+
+      - name: Set up Node.js
+        uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
+        with:
+          node-version: "20"
```
agriyakhetarpal:
```suggestion
          node-version: "24"
```

To match the other integration test job

.github/workflows/main.yml
```diff
@@ -64,6 +64,54 @@ jobs:
...
+          #include <stdio.h>
+          int main() {
+            printf("emcc works\n");
+            return 0;
+          }
+          "@
+          # Use ASCII encoding; PowerShell's default UTF-16 BOM is not supported by emcc
+          $code | Out-File -Encoding ASCII test.c
+          emcc test.c -o test.js
+          node test.js
```
agriyakhetarpal: Hmm, PowerShell troubles 😅

I wonder if we can make it default to UTF-8. Maybe something like this? I tried to search how to do so, and this is what showed up on Google's AI Mode

```powershell
    $PSDefaultParameterValues['*:Encoding'] = 'utf8'
    [Console]::OutputEncoding = [System.Text.UTF8Encoding]::new()

    # (rest of the commands)
```

See also https://github.com/PowerShell/PowerShell/issues/7233. Looks like it was closed by Microsoft but wasn't actually resolved and nothing happened :(

pyodide_build/build_env.py
```diff
@@ -329,23 +338,38 @@ def activate_emscripten_env(emsdk_dir: Path) -> dict[str, str]:
...
-    emsdk_env_script = emsdk_dir / "emsdk_env.sh"
+    emsdk_env_script_filename = "emsdk_env.bat" if IS_WIN else "emsdk_env.sh"
+    emsdk_env_script = emsdk_dir / emsdk_env_script_filename
     if not emsdk_env_script.exists():
-        raise FileNotFoundError(f"emsdk_env.sh not found at {emsdk_env_script}")
+        raise FileNotFoundError(
+            f"{emsdk_env_script_filename} not found at {emsdk_env_script}"
+        )
 
     # Source emsdk_env.sh and capture the resulting environment
```
agriyakhetarpal:
```suggestion
    # Source emsdk_env.bat/emsdk_env.sh and capture the resulting environment
```

agriyakhetarpal (commented):
Thanks! The Windows test looks great to me. However, `orjson` is failing to build on Linux because [the patches are being skipped](https://github.com/pyodide/pyodide-build/actions/runs/26025430202/job/76497427846?pr=322#step:9:47). 

Context: we had some trouble with Emscripten 4.0.9 building Rust side modules, which was fixed by https://github.com/emscripten-core/emscripten/pull/24427, merged into 4.0.12, and then we pulled it in here as a patch.

Here is a passing job on `main` which successfully applies the patches and builds `orjson`: [link](https://github.com/pyodide/pyodide-build/actions/runs/26016198043/job/76466657197#step:9:87).

pyodide_build/xbuildenv.py
```diff
@@ -540,20 +540,22 @@ def install_emscripten(
...
+                    cwd=emscripten_root,
+                    check=True,
+                )
```
agriyakhetarpal: Let's try this:

```suggestion
            for patch_file in sorted(patches_dir.glob("*.patch")):
                subprocess.run(
                    ["git", "apply", "--no-index", "--verbose", str(patch_file)],
                    cwd=emscripten_root,
                    check=True,
                )
```

.github/workflows/main.yml
Andrej730: Fixed in 7a8517c

.github/workflows/main.yml
Andrej730: Fixed in 733da18

---

Was testing if `[Console]::OutputEncoding = [System.Text.UTF8Encoding]::new()` would just do for it and it wasn't - powershell was requiring to provide `Out-File -Encoding` either explicitly or with `$PSDefaultParameterValues['*:Encoding'] = 'utf8'` and I thought - this can't be true in 2026 that everyone has to override it and I've checked powershell version CI was using and turned I was using old `powershell` instead of `pwsh`... which doesn't output utf 16 and it works. 😅

pyodide_build/build_env.py
Andrej730: Fixed in bb8ad70, found it in a few more places.

pyodide_build/xbuildenv.py
Andrej730: Fixed in aaaa6f5 and 53c62a4

Andrej730:
All green!

Sorry, for the noise, actual new commits are those 6:

image

agriyakhetarpal (approved):
Looks great to me now, thanks @Andrej730! I'll also request a review from Gyeongjae because he was wanting to review it in more detail.

ryanking13 (approved):
Thanks @Andrej730!

PR #336, opened 2026-04-27
Fix unisolated packages installation regarding pypa/build 1.4.4 update

ryanking13:
Related: https://github.com/pypa/build/pull/1040

pypa/build 1.4.4 introduced `--ignore-installed` flag when installing build dependencies. It basically overwrites all the pre-installed packages in the isolated environments and overwrites them with a fresh new package.

While that change makes sense, if breaks our build system as we symlink cross-compiled version of numpy and scipy.

This PR tries to fix it.

- [x] test this in pyodide/pyodide
- [x] changelog

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,12 +36,12 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
-          python-version: "3.14"
+          python-version: "3.12"  # TODO: update to 3.13
 
       - name: Set up Node.js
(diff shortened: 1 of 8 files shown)


agriyakhetarpal:
Argh. Thank you for taking it up. Maybe this is yet another sign we should probably vendor pypa/build...

hoodmane:
Let's just cap pypa build <= 1.4.3 for now.

henryiii:
I'm playing with an idea that might help in https://github.com/pypa/build/pull/1056.

henryiii:
This should be fixed now in build 1.5.0.

ryanking13:
Great! Thanks @henryiii!!

PR #337, opened 2026-04-27
Pin click version in tests temporarily

ryanking13:
Click 8.3.3 seems to have a regression breaking pytest. I was able to reproduce it locally but haven't got a simple reproducer that I can upstream yet. For now, pinning the click version so we can land other changes


ryanking13:
Reported upstream: https://github.com/pallets/click/issues/3384

PR #342, opened 2026-04-28
Pin build 1.4.3

hoodmane:
Workaround for https://github.com/pypa/build/pull/1040/ installing over our cross numpy and scipy.


ryanking13:
Thanks. I think `build~=1.4.0,<=1.4.3` would be better, but I think we should be able to remove the pin soon.

PR #343, opened 2026-04-29
Also accept legacy wheel when building

hoodmane:


ryanking13 (approved):
So I guess this allows passing prebuilt, legacy wheel in meta.yaml? Sounds okay to me.

PR #345, opened 2026-04-30
fix: loosen build requirement

henryiii:
Followup to #342, only 1.4.4 has an issue.


agriyakhetarpal (approved):

ryanking13:
Thank you so much @henryiii!

PR #347, opened 2026-05-18
Reliably find uv binary

agriyakhetarpal:
I made our uv helper match what cibuildwheel does to find uv. There's not much of a difference, but I've frequently run into issues with this when developing locally on my macOS machine and it might have been due to the functools cache for all we know.

Also, complies wth PEP 639 and fixes our licensing metadata.

### DIFF
--- a/pyodide_build/uv_helper.py
+++ b/pyodide_build/uv_helper.py
@@ -1,28 +1,20 @@
-import functools
-import os
+import contextlib
 import shutil
 
 
-@functools.cache
 def find_uv_bin() -> str | None:
     """
-    Check if the uv executable is available.
-
-    If the uv executable is available, return the path to the executable.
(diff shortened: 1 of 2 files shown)


ryanking13 (commented):
Thanks. So one difference this makes is that when there is `uv` installed, now we will always use uv, while previously, we used uv only when the pyodide entrypoint was invoked with uv (therefore the `UV` env variable is set).

I think it should be okay, but it would be worth to mention.

agriyakhetarpal:
Yes, correct. The errors I was facing were of the type where the cross-build packages were trying to be installed with `pip` in a `uv`-managed environment, but in such venvs the `pip` entry point does not exist deliberately and only `uv pip` works. I think it also makes expectations clearer, since we would expect that someone installing `pyodide-build[uv]` would want to use `uv`.

PR #348, opened 2026-05-18
MNT Replace a few more instances of bare `pip` calls with `sys.executable -m pip`

agriyakhetarpal:
I was browsing through what I had discussed on https://github.com/pyodide/pyodide-build/issues/58 and found a few more of these remaining.

I wonder if we should have an agentic style guide of sorts that can catch these kinds of things through custom instructions. I doubt Copilot reads `AGENTS.md` files; I can't get it to do so reliably across conversations.

### DIFF
--- a/pyodide_build/recipe/builder.py
+++ b/pyodide_build/recipe/builder.py
@@ -643,6 +643,8 @@ class RecipeBuilderPackage(RecipeBuilder):
             if self.build_metadata.cross_build_env:
                 subprocess.run(
                     [
+                        sys.executable,
+                        "-m",
                         "pip",
                         "install",
                         # Upgrade the package in the host environment if there is a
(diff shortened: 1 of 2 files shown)


pyodide_build/recipe/builder.py
```diff
@@ -643,6 +643,8 @@ def _package_wheel(
             if self.build_metadata.cross_build_env:
                 subprocess.run(
                     [
+                        sys.executable,
+                        "-m",
```
agriyakhetarpal:
```suggestion
```

pyodide_build/recipe/builder.py
```diff
@@ -666,6 +668,8 @@ def _package_wheel(
                 # overwriting the dependencies in the host environment.
                 subprocess.run(
                     [
+                        sys.executable,
+                        "-m",
```
agriyakhetarpal:
```suggestion
```

agriyakhetarpal:
I guess the only change that remains is in a test file 😅 That is not worth it. I'll close this!

PR #349, opened 2026-05-20
Accept `published_at` field for `CrossBuildEnvReleaseSpec` and add publish date information to `MetadataView`

agriyakhetarpal:
This is a companion PR to https://github.com/pyodide/pyodide/pull/6248. I'm trying to add metadata about when the cross-build environments are published, so consumers of such information can use it however they like.

- `CrossBuildEnvReleaseSpec` now takes in a `published_at` field with ISO 8601 timestamps
- MetadataView, i.e., the table and the JSON, have a "Published" column and a `published_at` field. They are currently blank and `null`, but will work once the corresponding Pyodide PR is merged and deployed to GitHub Pages, i.e., the `published_at` dates are included in the live JSON file.

### DIFF
--- a/pyodide_build/views.py
+++ b/pyodide_build/views.py
@@ -13,7 +13,7 @@ class MetadataView:
     emscripten: str
     pyodide_build: dict[str, str | None]
     compatible: bool
-    published_at: str | None = None
+    published_at: str = ""
 
     @classmethod
     def to_table(cls, views: list["MetadataView"]) -> str:
@@ -61,7 +61,7 @@ class MetadataView:
             pyodide_build_range = (
                 f"{view.pyodide_build['min'] or ''} - {view.pyodide_build['max'] or ''}"
             )
-            published = (view.published_at or "")[:10]
+            published = view.published_at[:10]
             row = [
                 f"{view.version:<{columns[0][1]}}",
                 f"{view.python:<{columns[1][1]}}",
--- a/pyodide_build/views.py
+++ b/pyodide_build/views.py
@@ -13,6 +13,7 @@ class MetadataView:
     emscripten: str
     pyodide_build: dict[str, str | None]
     compatible: bool
+    published_at: str | None = None
 
     @classmethod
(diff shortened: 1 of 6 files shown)


ryanking13:
Thanks. But this means the older versions of pyodide-build will suddenly stop working because of the remote metadata file change. I should've made pyodide-build robust on the spec changes...

ryanking13:
One workaround that I can think of is changing the metadata url, say, keeping the

```
https://pyodide.github.io/pyodide/api/pyodide-cross-build-environments.json
```

point to the old metadata file without the `publiched_at` key, freeze it.

and then, we make a new `v2` url with a new metadata file with the new keys

```
https://pyodide.github.io/pyodide/api/v2/pyodide-cross-build-environments.json
```

and update the url in pyodide-build (+ allow extra keys)

agriyakhetarpal:
Oh, thanks for raising this. Good idea. I think we should loosen our requirements for the spec going forward at the same time, so that we don't need to introduce a v3 unless we really really need it...

agriyakhetarpal:
I'll update this PR with the changes in a bit!

agriyakhetarpal:
Okay, we didn't exactly need a lot of changes here. I'm keeping this as draft until we have the v2 URL live on GitHub Pages.

pyodide_build/views.py
```diff
@@ -13,6 +13,7 @@ class MetadataView:
     emscripten: str
     pyodide_build: dict[str, str | None]
     compatible: bool
+    published_at: str | None = None
```
ryanking13: maybe set default value as empty string?

```suggestion
    published_at: str = ""
```

I think this is better than doing  `view.published_at or ""` when we need to print it

ryanking13 (approved):
Thanks! Let's merge this after we start publishing v2 metadata. I left a minor comment otherwise looks good.

pyodide_build/views.py
agriyakhetarpal: Yes, makes sense!

PR #350, opened 2026-05-21
Expose nightly and debug xbuildenvs in `pyodide xbuildenv search` and allow installing them via `pyodide xbuildenv install`

agriyakhetarpal:
This PR adds `--nightly` and `--debug` fields based on changes made in https://github.com/pyodide/pyodide-build-environment-nightly, particularly based on the PRs I set up there ([#72](https://github.com/pyodide/pyodide-build-environment-nightly/pull/72), [#73](https://github.com/pyodide/pyodide-build-environment-nightly/pull/73), [#74](https://github.com/pyodide/pyodide-build-environment-nightly/pull/74)). It is a follow-up to #349.

The fields are exclusive; i.e., `pyodide xbuildenv search` and `--all` work the same as before, but including `--nightly` or `--debug` only displays them. Here is a table to summarise the changes:

| Command | Information |
|--------|--------|
| `pyodide xbuildenv search` | Prints compatible stable xbuildenvs based on info in v2 spec |
| `pyodide xbuildenv search --all` | Prints both compatible and incompatible stable xbuildenvs based on info in v2 spec |
| `pyodide xbuildenv search --debug` | Prints compatible nightly-debug xbuildenvs. This can also include stable-debug xbuildenvs in the future |
| `pyodide xbuildenv search --nightly` | Prints compatible nightly-release xbuildenvs |
| `pyodide xbuildenv search --debug --nightly` | Prints compatible nightly-release and nightly-debug xbuildenvs | 
| `pyodide xbuildenv search -all --debug --nightly` | Prints both compatible and incompatible nightly-release and nightly-debug xbuildenvs |

Of course, the order of `--all`, `--debug`, and `--nightly` doesn't matter.

We can install the xbuildenvs, via `--nightly` and `--debug` flags to `pyodide xbuildenv install`, after finding them through the `search` flag.

As a result, we now have three URLs:
- Pyodide: https://pyodide.github.io/pyodide/api/v2/pyodide-cross-build-environments.json
- pyodide-build-environment-nightly: 
	- release builds: https://pyodide.github.io/pyodide-build-environment-nightly/api/v2/release.json
	- debug builds: https://pyodide.github.io/pyodide-build-environment-nightly/api/v2/debug.json

The base GitHub Pages page for pyodide-build-environment-nightly is a basic HTML page with some brief information to accompany these changes: https://pyodide.github.io/pyodide-build-environment-nightly. 

Additionally, I made two more changes:
- The length of the `MetadataView` columns is now computed dynamically based on the length of the items inside them, to prevent breaking the table due to overflows.
- For the pyodide-build version compatibility range, we now say `0.26.0 and later` instead of `0.26.0 -`, to be clearer about the ranges and mislead someone into thinking that the table is empty. The programmatic JSON output remains unchanged.

AI disclosure: I used Claude 4.5 Haiku to draft the tests.

Follow-ups we can do in here and Pyodide:
- [ ] Here: add the ABI version to the table, somehow, in an "ABI version" column.
- [x] Pyodide: publish stable-debug releases as well, to complete the matrix of possible combinations, since we do nightly-debug? I reckon it'd be pretty useful.

### DIFF
--- a/pyodide_build/cli/xbuildenv.py
+++ b/pyodide_build/cli/xbuildenv.py
@@ -94,26 +94,19 @@ def _install(
     Arguments:
         VERSION: version of cross-build environment to install (optional)
     """
-    manager = CrossBuildEnvManager(path)
-
-    if url:
-        manager.install(url=url, force_install=force_install)
-    elif nightly or debug:
-        nightly_url = (
+    if nightly or debug:
+        metadata_url = (
             NIGHTLY_DEBUG_CROSS_BUILD_ENV_METADATA_URL
             if debug
             else NIGHTLY_CROSS_BUILD_ENV_METADATA_URL
         )
-        metadata = load_cross_build_env_metadata(nightly_url)
-        if version:
-            release = metadata.get_release(version)
-        else:
-            releases = list(metadata.releases.values())
-            if not releases:
-                click.echo(f"No {'debug ' if debug else ''}nightly releases found.")
-                raise SystemExit(1)
-            release = releases[0]  # sorted newest-first
-        manager.install(url=release.url, force_install=force_install)
+    else:
+        metadata_url = None
+
+    manager = CrossBuildEnvManager(path, metadata_url=metadata_url)
+
+    if url:
+        manager.install(url=url, force_install=force_install)
     else:
         manager.install(version=version, force_install=force_install)
 
@@ -277,16 +270,19 @@ def _search(
             ),
         )
 
-    # Stable releases (always included)
-    stable_metadata = load_cross_build_env_metadata(
-        metadata_path or cross_build_env_metadata_url()
-    )
-    views = [
-        _make_view(r, "stable")
-        for r in stable_metadata.list_compatible_releases(**_compat_kwargs())
-    ]
+    views = []
 
-    # Nightly and/or debug releases (additive)
+    # Stable releases (only when neither --nightly nor --debug is passed)
+    if not (nightly or debug):
+        stable_metadata = load_cross_build_env_metadata(
+            metadata_path or cross_build_env_metadata_url()
+        )
+        views = [
+            _make_view(r, "stable")
+            for r in stable_metadata.list_compatible_releases(**_compat_kwargs())
+        ]
+
+    # Nightly and/or debug releases
     extra_sources = []
     if nightly:
         extra_sources.append(("nightly", NIGHTLY_CROSS_BUILD_ENV_METADATA_URL))
--- a/pyodide_build/cli/xbuildenv.py
+++ b/pyodide_build/cli/xbuildenv.py
@@ -7,6 +7,8 @@ from pyodide_build.common import default_xbuildenv_path
 from pyodide_build.views import MetadataView
 from pyodide_build.xbuildenv import CrossBuildEnvManager
 from pyodide_build.xbuildenv_releases import (
+    NIGHTLY_CROSS_BUILD_ENV_METADATA_URL,
+    NIGHTLY_DEBUG_CROSS_BUILD_ENV_METADATA_URL,
     cross_build_env_metadata_url,
     load_cross_build_env_metadata,
 )
@@ -52,6 +54,18 @@ def check_xbuildenv_root(path: Path) -> None:
     help="force installation even if the version is not compatible.",
 )
 @click.option(
+    "--nightly",
+    is_flag=True,
+    default=False,
+    help="install a nightly cross-build environment instead of a stable release.",
+)
+@click.option(
+    "--debug",
+    is_flag=True,
+    default=False,
+    help="install the debug variant of the cross-build environment (nightly only).",
+)
+@click.option(
     "--skip-cross-build-packages",
     is_flag=True,
     default=False,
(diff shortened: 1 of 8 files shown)


pyodide_build/cli/xbuildenv.py
```diff
@@ -81,15 +97,25 @@ def _install(
...
         )
+        metadata = load_cross_build_env_metadata(nightly_url)
+        if version:
+            release = metadata.get_release(version)
+        else:
+            releases = list(metadata.releases.values())
+            if not releases:
+                click.echo(f"No {'debug ' if debug else ''}nightly releases found.")
+                raise SystemExit(1)
+            release = releases[0]  # sorted newest-first
```
ryanking13: Can we hide this logic into `manager` class? It is not ideal to expose these logics into the cli function.

The manager class already have a logic to pick the best cross-build env when the metadata file is given, and it takes `metadata_url` as a parameter in the constructor.

I think it is okay to call `cross_build_env_metadata_url()` here and pass the returned URL into the manager, but I would like to avoid doing more than that.

pyodide_build/cli/xbuildenv.py
```diff
@@ -238,18 +270,49 @@ def _search(
...
             published_at=release.published_at,
+            source=source,
             compatible=release.is_compatible(
                 python_version=local["python"],
                 pyodide_build_version=local["pyodide-build"],
             ),
         )
-        for release in releases
+
+    # Stable releases (always included)
```
ryanking13: Is it the right behavior we want: showing nightly/debug versions mixed with the stable version?

I think people will be confused when they see the stable version listed with debug version, but it fails to install when they run `pyodide xbuildenv <stable_version> --debug`.

pyodide_build/views.py
```diff
@@ -14,16 +14,49 @@ class MetadataView:
     pyodide_build: dict[str, str | None]
     compatible: bool
     published_at: str = ""
+    source: str = field(
+        default="stable"
+    )  # "stable", "nightly", or "nightly-debug" # TODO: add stable-debug builds?
```
ryanking13, line 19: Yeah, probably we should release debug builds for stable versions. We already have that in our CDN, so we just need to update our CI a bit to include it in the distribution.

ryanking13 (commented):
Thanks! I am good with the overall direction, but added comments about some details.

pyodide_build/cli/xbuildenv.py
agriyakhetarpal: Hmm, you're right. Now that you mention this, I think this being additive won't be very useful. Let me rework it to being exclusive for nightly and debug.

pyodide_build/cli/xbuildenv.py
```diff
@@ -238,18 +263,52 @@ def _search(
...
+        stable_metadata = load_cross_build_env_metadata(
+            metadata_path or cross_build_env_metadata_url()
+        )
+        views = [
+            _make_view(r, "stable")
+            for r in stable_metadata.list_compatible_releases(**_compat_kwargs())
+        ]
+
+    # Nightly and/or debug releases
+    extra_sources = []
```
ryanking13: As we are now showing either stable or nightly/debug releases exclusively, I think we can cleanup this logic a bit (and rename `extra_sources`).

ryanking13 (approved):

pyodide_build/cli/xbuildenv.py
agriyakhetarpal: Ah, right. Yes, it's an intermediate list, and I think we can drop it entirely. Done in ede639043dfc9042bb4d27788ec662bd01f8a19e.

pyodide_build/cli/xbuildenv.py
```diff
@@ -238,18 +263,49 @@ def _search(
...
+            sources.append(("nightly", NIGHTLY_CROSS_BUILD_ENV_METADATA_URL))
+        if debug:
+            sources.append(
+                ("nightly-debug", NIGHTLY_DEBUG_CROSS_BUILD_ENV_METADATA_URL)
+            )
+        views = [
+            _make_view(r, source)
+            for source, url in sources
+            for r in load_cross_build_env_metadata(url).releases.values()
+            if show_all or r.is_compatible(**_compat_kwargs())
```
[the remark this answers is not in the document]
agriyakhetarpal: `list_compatible_releases` sorts with `packaging.version.Version`, which only works for PEP 440 versions, so we'll have to sort by `published_at` manually (which is fine, because they're all ISO 8601, so lexicographic sorting applies)

pyodide_build/views.py
```diff
@@ -14,16 +14,49 @@ class MetadataView:
...
-            ("Compatible", 10),
+    def to_table(cls, views: list["MetadataView"], show_source: bool = False) -> str:
+        # Build cell values first so we can measure them for column widths
+        rows: list[list[str]] = []
+        for view in views:
+            mn, mx = view.pyodide_build["min"], view.pyodide_build["max"]
+            if mn and mx:
+                pyodide_build_range = f"{mn} - {mx}"
+            elif mn:
+                pyodide_build_range = f"{mn} and later"
```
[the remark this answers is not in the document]
agriyakhetarpal: `max` without `min` never appears, and semantically it wouldn't make sense either, since a release always has a minimum compatible `pyodide-build` version (the one that introduced it)

PR #355, opened 2026-05-27
Automatically generate CLI reference documentation via `sphinx-click` and a custom directive

agriyakhetarpal:
I noticed in #350 that our [CLI reference page](https://pyodide-build.readthedocs.io/en/latest/reference/cli.html) is likely to get outdated often, as we may forget to update it with new CLI flags and subcommands being added. This PR introduces an inline Sphinx extension with a directive that scans entry points added by `pyodide-build` in `pyodide-cli` and adds them to the page. See the PR preview: https://pyodide-build--355.org.readthedocs.build/en/355/reference/cli.html

We can also skip such automation if we want, but based on my experience writing customisations for the Sphinx build process, this will remain fairly stable and shouldn't break in the future. It also beats adding new commands manually, so I recommend keeping it :D

The extension doesn't support Sphinx 9 yet, as I noticed from this issue that I encountered: https://github.com/click-contrib/sphinx-click/issues/159, which has a PR to fix it: https://github.com/click-contrib/sphinx-click/pull/160. We don't need any cutting-edge features from Sphinx anyway and none of our extensions imposes a Sphinx >=9 bound, so we should be fine with 8.x. Also, the extension doesn't support Markdown, but that's easily workaroundable with `{eval-rst}`: https://github.com/click-contrib/sphinx-click/issues/127.

Lastly, I dropped the `requirements-doc.txt` file, as it's kind of redundant with dependency groups now.

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -29,17 +29,15 @@ These keys can be set in `pyproject.toml` under `[tool.pyodide.build]` or via en
 
 | pyproject.toml key | Env variable | Description |
 |---|---|---|
-| `cflags` | `SIDE_MODULE_CFLAGS` | C compiler flags |
-| `cxxflags` | `SIDE_MODULE_CXXFLAGS` | C++ compiler flags |
-| `ldflags` | `SIDE_MODULE_LDFLAGS` | Linker flags |
+| `cflags` | `CFLAGS` | C compiler flags |
+| `cxxflags` | `CXXFLAGS` | C++ compiler flags |
+| `ldflags` | `LDFLAGS` | Linker flags |
 | `rustflags` | `RUSTFLAGS` | Rust compiler flags |
 | `rust_toolchain` | `RUST_TOOLCHAIN` | Required Rust nightly toolchain |
-| `rust_emscripten_target_url` | `RUST_EMSCRIPTEN_TARGET_URL` | URL for the Rust Emscripten target |
 | `meson_cross_file` | `MESON_CROSS_FILE` | Path to the Meson cross file |
 | `xbuildenv_path` | `PYODIDE_XBUILDENV_PATH` | Path to the cross-build environment |
 | `ignored_build_requirements` | `IGNORED_BUILD_REQUIREMENTS` | Space-separated PEP 508 build requirements to ignore |
 | `skip_emscripten_version_check` | `SKIP_EMSCRIPTEN_VERSION_CHECK` | Skip Emscripten version compatibility check (`0`/`1`) |
-| `build_dependency_index_url` | `BUILD_DEPENDENCY_INDEX_URL` | PyPI index URL for host-side build dependencies |
 | `default_cross_build_env_url` | `DEFAULT_CROSS_BUILD_ENV_URL` | URL override for the cross-build environment archive |
 | `use_legacy_platform` | `USE_LEGACY_PLATFORM` | Use the legacy `pyodide_*` platform tag instead of `pyemscripten_*` (`0`/`1`) |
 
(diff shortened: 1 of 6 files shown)


ryanking13 (commented):
Thanks! Would it be able to hide some of the subcommands from the generation? I intentionally not included `build-recipes`, `build-recipes-no-deps`, `skeleton` subcommands as I don't want to distract users with recipe-related commands. Most of the users who are looking at that document don't need to be aware of the recipe.

agriyakhetarpal:
Done! I also updated the config variables table since it was a bit outdated. Also, I'll open an issue to reorganise the documentation using Diátaxis.

docs/reference/configuration.md
```diff
@@ -25,18 +25,26 @@ xbuildenv_path = "/path/to/xbuildenv"
...
+These keys can be set in `pyproject.toml` under `[tool.pyodide.build]` or via environment variables:
 
 | pyproject.toml key | Env variable | Description |
 |---|---|---|
-| `cflags` | `CFLAGS` | C compiler flags |
-| `cxxflags` | `CXXFLAGS` | C++ compiler flags |
-| `ldflags` | `LDFLAGS` | Linker flags |
+| `cflags` | `SIDE_MODULE_CFLAGS` | C compiler flags |
+| `cxxflags` | `SIDE_MODULE_CXXFLAGS` | C++ compiler flags |
+| `ldflags` | `SIDE_MODULE_LDFLAGS` | Linker flags |
```
ryanking13: Let's revert these. In out-of-tree builds, CFLAGS, CXXFLAGS, and LDFLAGS work.

docs/reference/configuration.md
```diff
@@ -25,18 +25,26 @@ xbuildenv_path = "/path/to/xbuildenv"
...
 | `rustflags` | `RUSTFLAGS` | Rust compiler flags |
-| `rust_toolchain` | `RUST_TOOLCHAIN` | Rust nightly toolchain version |
-| `meson_cross_file` | `MESON_CROSS_FILE` | Path to Meson cross file |
-| `xbuildenv_path` | `PYODIDE_XBUILDENV_PATH` | Path to cross-build environment |
-| `ignored_build_requirements` | `IGNORED_BUILD_REQUIREMENTS` | Build requirements to ignore |
+| `rust_toolchain` | `RUST_TOOLCHAIN` | Required Rust nightly toolchain |
+| `rust_emscripten_target_url` | `RUST_EMSCRIPTEN_TARGET_URL` | URL for the Rust Emscripten target |
```
ryanking13: Let's remove this one. Most of the users will not use it.

docs/reference/configuration.md
```diff
@@ -25,18 +25,26 @@ xbuildenv_path = "/path/to/xbuildenv"
...
+| `meson_cross_file` | `MESON_CROSS_FILE` | Path to the Meson cross file |
+| `xbuildenv_path` | `PYODIDE_XBUILDENV_PATH` | Path to the cross-build environment |
+| `ignored_build_requirements` | `IGNORED_BUILD_REQUIREMENTS` | Space-separated PEP 508 build requirements to ignore |
+| `skip_emscripten_version_check` | `SKIP_EMSCRIPTEN_VERSION_CHECK` | Skip Emscripten version compatibility check (`0`/`1`) |
+| `build_dependency_index_url` | `BUILD_DEPENDENCY_INDEX_URL` | PyPI index URL for host-side build dependencies |
```
ryanking13: Let's remove this one too. I introduced it for #68, but it never reached production.

ryanking13 (approved):
Thanks! I left some minor comments, otherwise looks good to me

agriyakhetarpal:
Thanks! Done!

PR #360, opened 2026-06-03
Add correct Python version when generating lockfile

agriyakhetarpal:
The `info` block in our lockfiles was constructed with `"python": sys.version.partition(" ")[0]`, which is the host Python and not our cross-compiled CPython.

I noticed this when inspecting `pyodide-lock.json` included in https://github.com/pyodide/pyodide-build-environment-nightly/releases/tag/20260601, where the Python version is reported as 3.14.5 instead of 3.14.2. I did this while working on https://github.com/pyodide/pyodide-build-environment-nightly/pull/79.

So I guess the CI machine happened to have Python 3.14.5, while `PYVERSION` in `Makefile.envs` is the more correct value.

I am actually not very familiar with our pyodide-lock code to know if this is a consequential change or breaks any workflow there, so LMKWYT about this @ryanking13 – Thanks!

### DIFF
--- a/pyodide_build/recipe/graph_builder.py
+++ b/pyodide_build/recipe/graph_builder.py
@@ -904,7 +904,7 @@ def generate_lockfile(
         "arch": arch,
         "platform": platform,
         "version": build_env.get_build_flag("PYODIDE_VERSION"),
-        "python": sys.version.partition(" ")[0],
+        "python": build_env.get_build_flag("PYVERSION"),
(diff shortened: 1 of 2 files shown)


ryanking13 (approved):
Oh, yeah, this must be a bug. Thanks for catching it. Before we unvendor pyodide-recipes, the host version was exactly the same as the build version, so this was working by accident.

agriyakhetarpal:
Thanks for the review!

PR #362, opened 2026-06-05
Remove building from url, pypi, requirements features

ryanking13:
I would like to remove these features to reduce our maintenance burden. These features were requested and added by @joemarshall as a workaround before PEP 783 is accepted, so that people can build a set of packages that they want to use.

However, as pyodide wheels can be uploaded into PyPI, I don't want to maintain this feature. Instead, users should reach out to the package maintainers to build and upload wheels to PyPI.


agriyakhetarpal (commented):
I am +1 on this change, but we might want to think about this. We are still going to have some packages that retain patches, and removing them may cause regressions for users and towards our adoption, as was the case with the disabled packages landing in 0.28. Is there a way for us to estimate the impact? I don't think many people were using these features, as you mentioned, but it might still be good to do a sanity check by making this softer: deprecate with a notice in one release, then remove in the next?

I haven't had much of a use for these features before, so just to clarify, our feature to do nothing and just download the wheel if a wheel URL is specified in the recipe file is not going away here, and these are all just `pyodide build` features, not `pyodide build-recipes` features?

agriyakhetarpal (approved):
Oops, I meant to approve this alongside my question, not comment.

ryanking13:
Yeah, this does not affect`pyodide build-recipes`, only affect `pyodide build`. There is no reason to hurry, so we can wait for 314.0.0 release, start poking people to upload pyemscripten wheels, and then we can merge this PR.

agriyakhetarpal:
Makes sense! I added this label, inspired by cibuildwheel, so that we will remember.

PR #363, opened 2026-06-06
Add verbosity handling and pass through to pypa/build

### LINKED ISSUE #222: Support `pyodide build -v/vv`
## 🚀 Feature

When using `pyodide build` it would be nice to have a `-v` and `--vv` for verbose build logging

### Motivation

When using cibuildwheel to build pyodide wheels, a `build-verbosity` setting, https://cibuildwheel.pypa.io/en/stable/options/#build-verbosity, can be set which passes `-v` and `--vv` flags to the build step. Currently passing these flags to `pyodide build` will error e.g. https://github.com/pandas-dev/pandas/actions/runs/16574102747/job/46873985965?pr=61981

### Pitch

Add compatibility and any additional level of build verbosity to `pyodide build` by adding `-v` and `-vv` flags for interoperability with cibuildwheel and improved build context respectively.

### Alternatives

None

agriyakhetarpal:
I finally decided to take up #222. This PR exposes verbosity via `-v` and `-vv` CLI flags (stackable). They are passed along to pypa/build's `_ctx.VERBOSITY`. It controls pypa/build's own subprocess calls for uv/pip installs. Build backend commands (meson, cmake, gcc wrappers, etc.) are always visible regardless of verbosity.

Closes #222

TODO:
- [x] Update docs
- [x] Add some basic tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.34.5] - 2026/05/20
+## [0.35.0] - 2026/XX/XX
 
 ### Added
 
@@ -18,6 +18,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and `--debug` flags to search and install nightly and nightly-debug cross-build
   environments respectively.
   [#350](https://github.com/pyodide/pyodide-build/pull/350)
+
+- `pyodide build` now accepts `-v`/`--verbose` flags (stackable twice up to `-vv`)
+  to increase build verbosity. At `-v`, installer commands and
+  backend hook invocations are shown. At `-vv`, the package installers (`uv`/`pip`)
+  additionally receive a `-v` flag for detailed package resolution output.
+  [#363](https://github.com/pyodide/pyodide-build/pull/363)
 
 ## [0.34.4] - 2026/05/15
 
--- a/pyodide_build/pypabuild.py
+++ b/pyodide_build/pypabuild.py
@@ -165,6 +165,7 @@ def _build_in_isolated_env(
     outdir: str,
     distribution: Literal["sdist", "wheel"],
     config_settings: ConfigSettingsType,
+    verbosity: int = 0,
 ) -> str:
     # For debugging: The following line disables removal of the isolated venv.
     # It will be left in the /tmp folder and can be inspected or entered as
@@ -176,7 +177,7 @@ def _build_in_isolated_env(
         builder = ProjectBuilder.from_isolated_env(
             env,
             srcdir,
-            runner=_gen_runner(build_env, env),
+            runner=_gen_runner(build_env, env, verbosity=verbosity),
         )
 
         # first install the build dependencies
@@ -222,9 +223,10 @@ def _build_in_current_env(
     distribution: Literal["sdist", "wheel"],
     config_settings: ConfigSettingsType,
     skip_dependency_check: bool = False,
+    verbosity: int = 0,
 ) -> str:
     with common.replace_env(build_env):
-        builder = ProjectBuilder(srcdir, runner=_gen_runner(build_env))
+        builder = ProjectBuilder(srcdir, runner=_gen_runner(build_env, verbosity=verbosity))
 
         if not skip_dependency_check:
             missing = builder.check_dependencies(distribution, config_settings or {})
@@ -375,30 +377,44 @@ def build(
     config_settings: ConfigSettingsType,
     isolation: bool = True,
     skip_dependency_check: bool = False,
+    verbosity: int = 0,
 ) -> str:
-    try:
-        with _handle_build_error():
-            if isolation:
-                built = _build_in_isolated_env(
-                    build_env, srcdir, str(outdir), "wheel", config_settings
+    import logging
+
+    from build import _ctx as _build_ctx
+
+    _build_ctx.VERBOSITY.set(verbosity)
+    log_level = logging.DEBUG if verbosity >= 1 else logging.INFO
+    with set_log_level(logger, log_level):
+        try:
+            with _handle_build_error():
+                if isolation:
+                    built = _build_in_isolated_env(
+                        build_env,
+                        srcdir,
+                        str(outdir),
+                        "wheel",
+                        config_settings,
+                        verbosity=verbosity,
+                    )
+                else:
+                    built = _build_in_current_env(
+                        build_env,
+                        srcdir,
+                        str(outdir),
+                        "wheel",
+                        config_settings,
+                        skip_dependency_check,
+                        verbosity=verbosity,
+                    )
+                print(
+                    "{bold}{green}Successfully built {}{reset}".format(
+                        built, **_styles.get()
+                    )
                 )
-            else:
-                built = _build_in_current_env(
-                    build_env,
-                    srcdir,
-                    str(outdir),
-                    "wheel",
-                    config_settings,
-                    skip_dependency_check,
-                )
-            print(
-                "{bold}{green}Successfully built {}{reset}".format(
-                    built, **_styles.get()
-                )
-            )
-            return built
-    except Exception as e:  # pragma: no cover
-        tb = traceback.format_exc().strip("\n")
-        print("\n{dim}{}{reset}\n".format(tb, **_styles.get()))
-        _error(str(e))
-        sys.exit(1)
+                return built
+        except Exception as e:  # pragma: no cover
+            tb = traceback.format_exc().strip("\n")
+            print("\n{dim}{}{reset}\n".format(tb, **_styles.get()))
+            _error(str(e))
+            sys.exit(1)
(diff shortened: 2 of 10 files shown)


agriyakhetarpal:
I have set up the plumbing here but `pyodide build -v` and `pyodide build -vv` both return the same output to me, as the default no-verbose-flag build, so this needs an investigation. Maybe the logger is not set up correctly.

agriyakhetarpal:
The logger is indeed not set up correctly, as pypa/build uses print calls and it prints to `sys.stderr`, but we print to `sys.stdout`, so something about our Rich logger swallows it and prevents it from reaching the fd, and I don't know what.

pyodide_build/pypabuild.py
```diff
@@ -364,19 +371,55 @@ def get_build_env(
...
+def _make_pypa_build_logger() -> Callable[[str], None]:
+    """
+    Returns a logger function compatible with build._ctx.LOGGER.
+
+    pypa/build's default _log_default sends messages to logging.getLogger('build')
+    at INFO level, but that logger has no handlers and an effective level of WARNING,
+    so all output is silently dropped when we use pypa/build as a library.
+
+    We mirror pypa/build's CLI logger where everything goes to stderr.
+    """
```
agriyakhetarpal: We don't need a verbosity check here because `_ctx.run_subprocess()` only calls this logger with subprocess content when `_ctx.VERBOSITY > 0`, pypa/build handles the verbosity upstream.

CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.34.5] - 2026/05/20
+## [0.35.0] - 2026/XX/XX
```
agriyakhetarpal, line 8: Will update the date when I create a release next week. @ryanking13, would you like to include #362 in the release, too?

ryanking13: No #362 doesn't need to be merged soon. You can make a release after merging this.

CHANGELOG.md
```diff
@@ -19,6 +19,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   environments respectively.
   [#350](https://github.com/pyodide/pyodide-build/pull/350)
 
+- `pyodide build` now accepts `-v`/`--verbose` flags (stackable twice up to `-vv`)
+  to increase build verbosity. At `-v`, installer commands and
+  backend hook invocations are shown. At `-vv`, the package installers (`uv`/`pip`)
+  additionally receive a `-v` flag for detailed package resolution output.
+  [#363](https://github.com/pyodide/pyodide-build/pull/363)
```
ryanking13, line 26: sounds good

pyodide_build/tests/test_oot_build.py
```diff
@@ -21,15 +16,61 @@ def test_non_platformed_build(dummy_xbuildenv, tmp_path):
...
+    assert "Building wheel" in err, (
+        f"Expected 'Building wheel' step message at verbosity={verbosity}"
+    )
+
+    # The pyproject_hooks invocation is printed with "> " prefix to stderr only
+    # when verbosity >= 1 (mirrors pypa/build's "> cmd" formatting).
+    if verbosity >= 1:
+        assert "> " in err, (
+            f"Expected '> cmd' backend call line on stderr at verbosity={verbosity}"
+        )
```
ryanking13: This is a very weak assert. Can we actually check some actual log message that we expect?

pyodide_build/pypabuild.py
```diff
@@ -364,19 +371,55 @@ def get_build_env(
...
     build_env: Mapping[str, str],
     config_settings: ConfigSettingsType,
     isolation: bool = True,
     skip_dependency_check: bool = False,
+    verbosity: int = 0,
 ) -> str:
+    from build import _ctx as _build_ctx
+
+    token_verbosity = _build_ctx.VERBOSITY.set(verbosity)
+    token_logger = _build_ctx.LOGGER.set(_make_pypa_build_logger())
```
ryanking13: Could you move this to `vendor/_pypabuild.py`? I would like to encapsulate any imports that use the private pypa/build APIs there.

pyodide_build/pypabuild.py
```diff
@@ -364,19 +371,55 @@ def get_build_env(
         yield env
 
 
+def _make_pypa_build_logger() -> Callable[[str], None]:
```
ryanking13: Maybe this can be moved to `vendor/_pypabuild` as well.

pyodide_build/pypabuild.py
ryanking13: I would say we should still check the verbosity, as the behavior you mentioned is an implementation detail that can change anytime.

pyodide_build/pypabuild.py
```diff
@@ -364,19 +371,55 @@ def get_build_env(
...
     outdir: Path,
     build_env: Mapping[str, str],
     config_settings: ConfigSettingsType,
     isolation: bool = True,
     skip_dependency_check: bool = False,
+    verbosity: int = 0,
 ) -> str:
+    from build import _ctx as _build_ctx
+
+    token_verbosity = _build_ctx.VERBOSITY.set(verbosity)
```
ryanking13: just curious, why do you call it "token"?

pyodide_build/pypabuild.py
```diff
@@ -364,19 +371,55 @@ def get_build_env(
...
+    Returns a logger function compatible with build._ctx.LOGGER.
+
+    pypa/build's default _log_default sends messages to logging.getLogger('build')
+    at INFO level, but that logger has no handlers and an effective level of WARNING,
+    so all output is silently dropped when we use pypa/build as a library.
+
+    We mirror pypa/build's CLI logger where everything goes to stderr.
+    """
+
+    def _log(message: str, *, kind: tuple[str, ...] | None = None) -> None:
```
ryanking13, line 399: Could you add a link to the original logger implementation in the pypa/build for a future reference? Also, adding a comment about what the message and kind are would help understand the code.

pyodide_build/pypabuild.py
```diff
@@ -186,7 +190,7 @@ def _build_in_isolated_env(
         except BuildBackendException:
             pass
 
-        if not build_reqs:
+        if build_reqs is None:
```
ryanking13, line 194: Is this change related to the logger? I believe this should be no harm, but let's separate the PR just in case.

ryanking13 (approved):
Thanks for handling this. I left some comments but overall it looks good to me

pyodide_build/pypabuild.py
agriyakhetarpal: Hmm, I'd like to keep this one here itself, as it's something that we are implementing (even though we keep it private). It is not vendored from pypa/build.

pyodide_build/pypabuild.py
agriyakhetarpal: Oh, that is because `ContextVars.set()` method returns a `Token` object. See https://docs.python.org/3/library/contextvars.html#contextvars.Token.

I think we could also flip it and call it `verbosity_token` and `logger_token`, as that might sound a bit nicer, but either of the options is clear to me :D

pyodide_build/pypabuild.py
agriyakhetarpal: No, it is a pre-existing bug that I caught due to the logger implementation. I added this in this commit: dda7ee5de6e332062066e60c4f92bbf95aaa2f2f

The idea is that both `None` and `set()` are falsy, so `if not build_reqs:` is `True` for both. `None` is the case where the PEP 517 `get_requires_for_build_wheel` build hook fails. And `set()` is the case where it passes, but has no dynamic build-time dependencies defined by it (for example, this is what scikit-build-core does to plug CMake/Ninja in if they are needed and are not already installed: https://github.com/scikit-build/scikit-build-core/blob/ed95f7bd499e183686b5f4b1c5c5fc3ba4507b06/src/scikit_build_core/setuptools/build_meta.py#L86-L95). Very few packages need this (NumPy does, for CPU detection I think), and most packages don't define it, so we get `set()` if the hook succeeds, but both of those can trigger the condition, and we get a double message as each call logs one message.

I also think it will be better to split it into its own PR. Thanks!

pyodide_build/pypabuild.py
agriyakhetarpal: Done!

pyodide_build/pypabuild.py
agriyakhetarpal: Yes, makes sense. Done here: b265166178975910a0532103bddf629cefd18dfe

pyodide_build/tests/test_oot_build.py
agriyakhetarpal: Agreed. I think 15d257329ff35eefd5a365571d0fbb870156cba3 should be better!

ryanking13 (approved):
Thanks!

pyodide_build/pypabuild.py
ryanking13: Sure, that's fine.

agriyakhetarpal:
Thanks for reviewing!

PR #364, opened 2026-06-06
Avoid double logging of PEP 517 `get_requires_for_build_wheel` build hook

agriyakhetarpal:
See explanation in https://github.com/pyodide/pyodide-build/pull/363#discussion_r3368207822. Split off from #363.


ryanking13 (approved):
Thanks!

PR #366, opened 2026-06-07
Ensure installation of cross-build packages before the build script runs

### LINKED ISSUE #365: Lazily installing cross-build packages is currently not reliable
With #297 and #302, we added a feature to install cross-build packages only when we require them. However, as part of debugging the implementation devised in #21 and testing it in https://github.com/pyodide/pyodide-recipes/pull/591, I think I have exposed a latent bug that has existed for a while, but our infrastructure has kept it from manifesting in CI builds.

To reproduce:

1. Set up the pyodide-recipes repository
2. Set up submodule to point to #21 (it is what exposes the bug right now, but the changes are not related)
3. Run `pyodide build-recipes RobotRaconteur` in the pyodide-recipes repository, fresh, before building anything else.

Right now, `ensure_cross_build_packages_installed()` on `main` is called inside `_build_in_isolated_env()`:

https://github.com/pyodide/pyodide-build/blob/29461276793f5efe59a31425f6e10b36777a2311/pyodide_build/pypabuild.py#L122-L123

https://github.com/pyodide/pyodide-build/blob/29461276793f5efe59a31425f6e10b36777a2311/pyodide_build/pypabuild.py#L183-L184

Which is part of `_compile()`. Now, looking at `_build_package` at https://github.com/pyodide/pyodide-build/blob/29461276793f5efe59a31425f6e10b36777a2311/pyodide_build/recipe/builder.py#L564-L584

The order is `bash_runner.run(self.build_metadata.script, ...)`, which runs the CMake configuration, and then this is followed by `self._compile(bash_runner)`, which is where we lazily load cross-build packages.

In this case, the RobotRaconteur recipe has a script section that needs `host_site_packages` populated before CMake runs, but `host_site_packages/numpy/_core/include` is still empty at execution time, so headers such as `ndarrayobject.h` are not found because they do not exist yet. There is no `.cross-build-packages-installed` marker either. Slightly before the last few commits I pushed in #21, we were not installing the cross-build packages; instead, we were only installing the cross-build files. `ndarrayobject.h` and other similar headers are not included in our cross-build files. 

Why this has never showed up for us is because by the time the build order reaches RobotRaconteur, some earlier package that lists `numpy` as a host dependency (could be `scipy`, `pandas`, or any other scientific package, etc.), has already necessitated the existence of the `_compile()` step and _warmed up_ the `host_site_packages` directory, which is then used by all the packages.

Building RobotRaconteur in isolation (just the recipe and its direct dependencies) does not trigger the lazy installation, so nothing is present in the cross-build packages.

Therefore, a fix could be to change when the lazy trigger runs, so that we detect the need to install, and install, the cross-build packages at the start of a package build rather than at a different event (if the package's recipe lists them in the `host:` section, of course).

agriyakhetarpal:
Closes #365

In our builder code, I've added a new method called `_ensure_cross_build_packages_for_host_requirements` to the base `RecipeBuilder`, and I am calling it before `self._build_package(bash_runner)` runs.

The idea is that we need to check whether the package's declared host requirements, say, `numpy` for RobotRaconteur) intersect with `get_unisolated_packages()`, and if so (and we are in an xbuildenv), we call our lazy-install machinery established via #302 a little earlier, to ensure that they get installed and the cross-build files are available when the build script runs.

I don't think I've found a better way to do this, as our build pipeline is pretty tricky to understand and test in isolation anyway.

There is an explanation in #365 for where this bug comes up in practice.

### DIFF
--- a/pyodide_build/recipe/builder.py
+++ b/pyodide_build/recipe/builder.py
@@ -291,6 +291,13 @@ class RecipeBuilder:
         ):
             self._build_package(bash_runner)
 
+    # Some packages such as RobotRaconteur and opencv-python reach into
+    # $HOSTINSTALLDIR/host site-packages directly from their build scripts
+    # which run before the isolated build environment is set up. The lazy
+    # install normally triggered from there would be too late for them, so
+    # we trigger it here instead, based on the package's declared host
+    # requirements. See https://github.com/pyodide/pyodide-build/issues/365
+    # for reference.
     def _ensure_cross_build_packages_for_host_requirements(self) -> None:
         """
         Make sure cross-build packages (numpy, scipy, etc.) are installed into
--- a/pyodide_build/recipe/builder.py
+++ b/pyodide_build/recipe/builder.py
@@ -25,9 +25,12 @@ from pyodide_build.build_env import (
     _create_constraints_file,
     get_build_environment_vars,
     get_build_flag,
+    get_current_xbuildenv_manager,
     get_pyodide_root,
     get_pyversion_major,
     get_pyversion_minor,
+    get_unisolated_packages,
+    in_xbuildenv,
     pyodide_tags,
     replace_so_abi_tags,
     wheel_platform,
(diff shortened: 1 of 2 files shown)


ryanking13 (approved):
Thanks. I am okay with this as a workaround, but we should probably upstream the problematic packages and let them use unisolated build instead.

pyodide_build/recipe/builder.py
```diff
@@ -280,12 +283,36 @@ def _build(self) -> None:
...
+        requirements need them.
+        """
+        if not in_xbuildenv():
+            return
+
+        unisolated_packages = set(get_unisolated_packages())
+        host_requirements = {req.lower() for req in self.recipe.requirements.host}
+
+        if unisolated_packages & host_requirements:
+            get_current_xbuildenv_manager().ensure_cross_build_packages_installed()
```
ryanking13, line 314: Hmm, wait. Actually, when building recipes, we don't rely on `ensure_cross_build_packages_installed` at all.

After building cross-build recipe such as numpy, it should install [into the host_site_package](https://github.com/pyodide/pyodide-build/blob/3806ca0a5be4b08f97c8380d93f86fe4da8f38b5/pyodide_build/recipe/builder.py#L670-L709) as a post build step, so it should be already installed there.

ryanking13 (changes requested):

PR #375, opened 2026-06-12
Resolve sys.prefix symlinks when replacing host include paths

### LINKED ISSUE #374: pyodide build uses host Python headers on macOS with homebrew Python
Building an out-of-tree package on macOS with a homebrew-installed Python fails during compilation with:

```
pyport.h:429:2: error: "LONG_BIT definition appears wrong for platform (bad gcc/glibc config?)."
```

because the host Python's include directory survives into the emscripten compile (wasm32 has 32-bit longs, while the host headers assume 64-bit).

**Root cause:** 

* `replay_genargs_handle_dashI` (`pywasmcross.py`) does not resolve symlinks in `sys.prefix`.

* Homebrew's prefix goes through a symlink:
```
/opt/homebrew/opt/python@3.14 -> ../Cellar/python@3.14/...
```

and homebrew Python reports the symlinked path as `sys.base_prefix` even when invoked via its realpath. 

* So the resolved include path (`Cellar`) never starts with the prefix (`opt`), and the header rewrite is skipped.

Reproduced with pyodide-build 0.35.0 / Pyodide 314.0.0 / homebrew Python 3.14.3, while building imgui-bundle's `pyemscripten` wheel out-of-tree.

I will open a PR with a fix (resolve both sides of the comparison) and a regression test.

pthom:
Fixes #374.

`replay_genargs_handle_dashI` now resolves symlinks for both include_path and sys.prefix before comparison.

The added regression test (real symlink in `tmp_path`) fails on `main` and passes with this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- Fixed an include path issue on macOS when homebrew Python is installed:
+  if `sys.prefix` contains a symlink (as is the case with homebrew), resolve it
+  so that host include paths are still replaced by the cross-compile ones
+  [#375](https://github.com/pyodide/pyodide-build/pull/375)
+
 ## [0.35.0] - 2026/06/09
 
 ### Added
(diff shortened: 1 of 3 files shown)


agriyakhetarpal (approved):
Nice, thanks @pthom! I was also having this problem with Homebrew Python on my machine, but I didn't spend enough time debugging!

Would you like me to cut a release with this change?

pthom:
Hi @agriyakhetarpal 

> Would you like me to cut a release with this change?

If possible, yes. At the moment, I do have to patch pyodide-build manually.

agriyakhetarpal:
Sorry for the trouble. Let me do so, then!

pthom:
Thank you very much for the quick release

Issue #376, opened 2026-06-12
Pyodide-build review from Claude Fable


henryiii:
:robot: _AI text below_ :robot:

All four reviews are complete and I've spot-checked the highest-impact claims directly in the source — they hold up. Here's the consolidated review.

# Project review: pyodide-build (at `e362d0a`, 0.35.1)

**TLDR:** The codebase is in good shape overall — the agents specifically verified that archive extraction is traversal-protected, checksums are enforced, the dependency graph is race-free, and typing/ruff hygiene is already modern. But the review found one high-severity hang, a cluster of real medium-severity bugs (several reproduced live), two big performance wins in the PyPI dependency resolver, and a handful of easy cleanups.

### High

- **Build hangs forever if a recipe script ends with `exit 0`** — `pyodide_build/recipe/bash_runner.py:52-62`. `run_unchecked` appends an env-dump command after the user's script. If the script terminates the shell itself (`exit 0`, or `exec` of a program that exits 0), bash returns rc 0 without ever writing to the pipe, and the parent blocks on `self._reader.readline()` indefinitely — there's no EOF because the parent holds the write fd open. The reviewing agent reproduced the hang with `run_unchecked('echo hi; exit 0')`.

### Medium

- **Cached build-env dict is mutated per package** — `pyodide_build/recipe/bash_runner.py:109-110`. `get_build_environment_vars()` is `@functools.cache`d and returns the cached dict itself; `env.update(extra_envs)` permanently injects per-package vars (`PKGDIR`, `PKG_VERSION`, `DISTDIR`, plus all of `os.environ`) into the cached value. When one process builds several packages sequentially (`pyodide build-recipes-no-deps pkg1 pkg2`), later packages see the previous package's values. Fix is one character class: `env = dict(get_build_environment_vars(pyodide_root))`.

- **`continue` should be `break` in wheel matching** — `pyodide_build/common.py:145-148`. A wheel whose filename carries a compressed tag set matching two supported tags is yielded once per tag, so `find_matching_wheel` raises `RuntimeError("Found multiple matching wheels")` listing the same file twice. This is plausible right now given the pyodide→pyemscripten platform-tag migration (`pyodide_tags()` includes both). Reproduced live with a single wheel on disk.

[...]

PR #377, opened 2026-06-12
fix: correct three bugs in pyodide venv CLI and cleanup logic

henryiii:

Part of #376

## Summary

This PR fixes three verified bugs in the `pyodide venv` subsystem:

**Bug 1 (medium): `--no-download` / `--never-download` were silent no-ops**

In `cli/venv.py`, the click option `--download/--no-download` (a boolean pair) was registered _after_ the `--no-download/--never-download` flag option. Click resolves `--no-download` to the later registration, so passing it set `download=False` (already the default) while `no_download` stayed `False` — nothing was forwarded to virtualenv. Fixed by replacing the boolean pair with a standalone `--download` `is_flag` option, leaving `--no-download` solely owned by the existing flag option.

**Bug 2 (low): versioned `pipX.Y` scripts vanish from created venvs**

In `out_of_tree/venv.py`, the loop that replaces pip scripts with symlinks computed the canonical path using `pip.with_suffix(self.exe_suffix)`. On Unix `exe_suffix` is `""`, so `Path("pip3.12").with_suffix("")` yields `pip3` — the `.12` is treated as a suffix. The fix introduces a `_pip_script_name(pip, exe_suffix)` helper that uses `str.removesuffix` so only a real suffix (e.g. `.bat`) is stripped and the full versioned name is preserved.

**Bug 3 (low): failure cleanup deletes a pre-existing destination directory**

In `out_of_tree/venv.py`, the `except` clause in `create()` called `shutil.rmtree(session.creator.dest)` unconditionally. If the user pointed `pyodide venv` at an existing directory and creation failed (e.g. network error during stdlib install), their directory was silently deleted. The fix records `dest_existed_before = dest_path.exists()` before calling `session.run()`, and on failure either leaves the directory intact with a warning (pre-existing) or removes it (newly created).

## Tests

New unit tests added to `pyodide_build/tests/test_venv.py`:

- `test_venv_cli_download_flags` — parametrized CliRunner test verifying that `--no-download`, `--never-download`, and `--download` each forward the correct argument to virtualenv (and the empty case forwards nothing).
- `test_pip_script_name_helper` — parametrized unit test for `_pip_script_name` covering Unix (no suffix) and Windows (`.bat` suffix) cases, including versioned names like `pip3.12`.
- `test_cleanup_skips_preexisting_directory` — monkeypatches `configure_virtualenv` to raise, verifies that the pre-existing destination directory and its contents are preserved after the failure.

## Test plan

- [x] `uv run --group test pytest pyodide_build/tests/test_venv.py -m "not integration" -v` — 21 passed
- [x] `uv run --group test pytest pyodide_build -m "not integration"` — 328 passed (4 pre-existing failures unrelated to this PR)
- [x] `prek -a --quiet` — passes

### DIFF
--- a/pyodide_build/out_of_tree/venv.py
+++ b/pyodide_build/out_of_tree/venv.py
@@ -52,6 +52,20 @@ def dedent(s: str) -> str:
     return textwrap.dedent(s).strip() + "\n"
 
 
+def _pip_script_name(pip: Path, exe_suffix: str) -> Path:
+    """Return the pip script path with *exe_suffix* applied.
+
+    On Unix ``exe_suffix`` is empty and the name is returned unchanged, so
+    versioned names like ``pip3.12`` keep their version (``Path.with_suffix("")``
+    would wrongly strip it to ``pip3``). On Windows the real extension (``.exe``
+    or ``.bat``) is replaced with ``exe_suffix``.
+    """
+    if not exe_suffix:
+        return pip
+    base_name = pip.stem if pip.suffix else pip.name
+    return pip.parent / (base_name + exe_suffix)
+
+
 def get_pyversion() -> str:
     return f"{sys.version_info.major}.{sys.version_info.minor}"
 
@@ -490,7 +504,7 @@ class PyodideVenv(ABC):
                 continue
             pip.unlink(missing_ok=True)
 
-            patched_pip_exe = pip.with_suffix(self.exe_suffix)
+            patched_pip_exe = _pip_script_name(pip, self.exe_suffix)
             if patched_pip_exe != self.pip_patched_path:
                 patched_pip_exe.unlink(missing_ok=True)
                 patched_pip_exe.symlink_to(self.pip_patched_path)
@@ -589,15 +603,24 @@ class PyodideVenv(ABC):
         session = self._create_session()
         self._pyodide_pyversion = check_host_python_version(session)
 
+        dest_path = Path(session.creator.dest)
+        dest_existed_before = dest_path.exists()
+
         try:
             session.run()
-            self._venv_root = Path(session.creator.dest).absolute()
+            self._venv_root = dest_path.absolute()
             self._venv_bin = self._venv_root / self.bin_dir_name
 
             self.configure_virtualenv()
             self._install_stdlib()
         except (Exception, KeyboardInterrupt, SystemExit):
-            shutil.rmtree(session.creator.dest)
+            if dest_existed_before:
+                logger.warning(
+                    "Venv creation failed; leaving pre-existing directory %s intact",
+                    dest_path,
+                )
+            else:
+                shutil.rmtree(dest_path, ignore_errors=True)
             raise
 
         logger.success("Successfully created Pyodide virtual environment!")
--- a/pyodide_build/tests/test_venv.py
+++ b/pyodide_build/tests/test_venv.py
@@ -8,8 +8,11 @@ from pathlib import Path
 from textwrap import dedent
 
 import pytest
+from click.testing import CliRunner
 
+from pyodide_build.cli import venv as venv_cli
 from pyodide_build.out_of_tree import app_data, venv
+from pyodide_build.out_of_tree.venv import _pip_script_name
 from pyodide_build.xbuildenv import CrossBuildEnvManager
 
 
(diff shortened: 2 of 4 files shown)


pyodide_build/out_of_tree/venv.py
```diff
@@ -42,6 +42,19 @@ def dedent(s: str) -> str:
...
+    """Return the canonical pip script path after stripping any existing suffix
+    and re-applying *exe_suffix*.
+
+    On Unix, ``exe_suffix`` is the empty string and ``Path.with_suffix("")``
+    would incorrectly strip the version component from names like ``pip3.12``
+    (producing ``pip3`` instead).  We therefore use ``str.removesuffix`` so
+    that only a real suffix — e.g. ``.bat`` on Windows — is removed.
+    """
+    base_name = pip.name.removesuffix(exe_suffix) if exe_suffix else pip.name
+    return pip.parent / (base_name + exe_suffix)
```
agriyakhetarpal: This would introduce a new regression on Windows, as it trades one bug for another.

While this fixes the `.12` mangling with the old code, pip scripts on Windows are `pip3.12.exe` and not `.bat`, so `removesuffix(".bat")` does nothing, and we end up appending `.bat` anyway, producing `pip3.12.exe.bat` (and that doesn't make sense)

```suggestion
    if not exe_suffix:
        return pip
    base_name = pip.stem if pip.suffix else pip.name
    return pip.parent / (base_name + exe_suffix)
```

pyodide_build/tests/test_venv.py
```diff
@@ -367,3 +370,146 @@ def test_create_app_data_dir(tmp_path, clear_app_data_cache):
...
+        # Unix: no suffix — versioned name must be preserved as-is
+        ("pip3.12", "", "pip3.12"),
+        ("pip3", "", "pip3"),
+        ("pip", "", "pip"),
+        # Windows: .bat suffix must be replaced without eating the version part
+        ("pip3.12.bat", ".bat", "pip3.12.bat"),
+        ("pip3.bat", ".bat", "pip3.bat"),
+        ("pip.bat", ".bat", "pip.bat"),
+    ],
+)
```
agriyakhetarpal, line 456:
```suggestion
@pytest.mark.parametrize(
    "script_name,exe_suffix,expected_name",
    [
        # Unix: no suffix — versioned name must be preserved as-is
        ("pip3.12", "", "pip3.12"),
        ("pip3", "", "pip3"),
        ("pip", "", "pip"),
        # Windows exe executables should be replaced with .bat
        ("pip3.12.exe", ".bat", "pip3.12.bat"),
        ("pip3.exe", ".bat", "pip3.bat"),
        ("pip.exe", ".bat", "pip.bat"),
        # Windows: already .bat input should be idempotent
        ("pip3.12.bat", ".bat", "pip3.12.bat"),
        ("pip3.bat", ".bat", "pip3.bat"),
        ("pip.bat", ".bat", "pip.bat"),
    ],
)
```

agriyakhetarpal (changes requested):
Bugs 1 and 3 make sense to me. Thanks @henryiii! Bug 2 needs more work.

Could you please clear some of the verbose docstrings in the tests as well? I don't think we should explicitly write "Bug 2" and "Bug 3" in the tests as we won't remember what those were later.

henryiii:
Claude loves to be verbose. :) Sure, will do this in a few minutes.

agriyakhetarpal:
Awesome, thanks! I find it very difficult to make it respect what I have in `~/.claude/CLAUDE.md`.

henryiii:
It always does some of that. Things like "be concise" though don't seem to work there. But "Prefix PR descriptions and comments on PRs with the line ":robot: _AI text below_ :robot:" to indicate you are an agent speaking on a user's behalf." has worked 100% of the time.

pyodide_build/out_of_tree/venv.py
henryiii: :robot: _AI text below_ :robot:

Good catch — fixed in c6306c4. `_pip_script_name` now returns early on Unix (empty suffix) and on Windows replaces the real extension via `pip.stem`, so `pip3.12.exe` → `pip3.12.bat` instead of `pip3.12.exe.bat`. Added `.exe` cases to the parametrized test.

pyodide_build/tests/test_venv.py
henryiii: :robot: _AI text below_ :robot:

Applied — the test now covers the `.exe` → `.bat` and idempotent-`.bat` cases.

henryiii:

Addressed the review and rebased onto `main` (force-pushed `c6306c4`):

- **Bug 2 (Windows):** reworked `_pip_script_name` per your suggestion so it no longer produces `pip3.12.exe.bat`; added the `.exe` cases to the parametrized test.
- **Test verbosity:** dropped the "Bug 1/2/3" labels and the long docstrings/comments; assertions are now plain.
- **CHANGELOG:** collapsed the three entries into a single concise line under `0.36.0`.

`prek -a` is clean and `test_venv.py` passes (35 passed, 1 pre-existing skip).

agriyakhetarpal (approved):
All good now, thanks for updating it so quickly!

PR #378, opened 2026-06-12
Fix cross-build-environment lifecycle bugs

henryiii:

Part of #376.

This PR fixes five verified bugs in the cross-build-environment (xbuildenv) lifecycle in `pyodide_build/xbuildenv.py`.

### Bug 1 — Python-version compatibility check was self-defeating
On a version-marker mismatch, `build_env._init_xbuild_env` calls `install()` and re-checks. But `install()` skipped all work when the resolved version directory already existed, then unconditionally called `_add_version_marker()`, rewriting the marker with the *current* Python version. The re-check then passed and builds proceeded with `HOSTSITEPACKAGES` contents installed under the *old* Python (the cross-build-packages marker still existed so they were never refreshed), and the intended "please switch back / reinstall" error was unreachable.

**Fix:** `install()` now detects a marker mismatch on an already-installed env (`_version_marker_matches_for(download_path)`), treats it as installation work, and: refreshes host packages (or drops the cross-build-packages marker so they're reinstalled lazily) and rewrites the version marker. The marker is now written **only when installation work was actually performed**, so a skipped install can no longer silently mask a mismatch.

### Bug 2 — broken package index when installing via `default_cross_build_env_url`
The explicit-`url` branch correctly skipped `_create_package_index` (the "version" is a mangled URL string). The `default_url` branch left `url=None`, so the index was created with hrefs like `CDN_BASE.format(version="https_example_com_...")`, which `pyodide venv` pip installs later consume.

**Fix:** both URL branches now set an `installed_from_url` flag, and index creation is skipped for either.

### Bug 3 — `use_version()` crashed on a dangling `xbuildenv` symlink
`if symlink_dir.exists()` follows the link and is `False` for a dangling symlink, so the stale link was never unlinked and `symlink_to()` raised `FileExistsError` on every subsequent install/use.

**Fix:** use `is_symlink()` (lexists semantics) so a dangling symlink is removed first.

### Bug 4 — failure cleanup deleted a pre-existing valid installation
The `except` block ran `shutil.rmtree(download_path)` even when `download_path` existed before this call. Any later failure (e.g. `symlink_to` raising `OSError` on Windows without symlink privilege) wiped the previously-good cached xbuildenv, and the `rmtree` could mask the original exception.

**Fix:** track whether *this* call created the directory and only delete it then; use `shutil.rmtree(..., ignore_errors=True)` so cleanup never masks the original error.

### Bug 5 — `_find_latest_version` mismatch diagnostics: wrong order + `IndexError`
When no compatible release existed, `python_versions` was built from releases sorted by *Pyodide* version, then indexed `[0]`/`[-1]` as if sorted by *Python* version, so newest/oldest messages could be wrong. An empty releases dict raised an uncaught `IndexError`.

**Fix:** sort the distinct Python versions by `packaging.version.Version`, derive newest/oldest from that, and raise a clear error for empty metadata.

### Tests
Added to `pyodide_build/tests/test_xbuildenv.py`:
- `test_use_version_dangling_symlink` — dangling-symlink recovery
- `test_install_preexisting_not_deleted_on_failure` / `test_install_new_deleted_on_failure` — pre-existing install preserved on failure; new install cleaned up
- `test_install_url_default_no_mangled_package_index` — default-URL install creates no mangled package index
- `test_install_version_marker_mismatch_triggers_reinstall` / `test_install_version_marker_match_no_marker_rewrite` — marker mismatch reinstalls and refreshes; matching marker is a no-op
- `test_find_latest_version_empty_releases` — clear error instead of `IndexError`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed several bugs in the cross-build-environment (xbuildenv) lifecycle
-  [#376](https://github.com/pyodide/pyodide-build/issues/376):
+  [#376](https://github.com/pyodide/pyodide-build/issues/376),
+  [#378](https://github.com/pyodide/pyodide-build/pull/378):
   - A Python-version marker mismatch on an already-installed xbuildenv no longer
     silently overwrites the marker with the current Python version. A reinstall
     now actually refreshes the host packages (or surfaces the intended error)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,29 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- Fixed several bugs in the cross-build-environment (xbuildenv) lifecycle
+  [#376](https://github.com/pyodide/pyodide-build/issues/376):
+  - A Python-version marker mismatch on an already-installed xbuildenv no longer
+    silently overwrites the marker with the current Python version. A reinstall
+    now actually refreshes the host packages (or surfaces the intended error)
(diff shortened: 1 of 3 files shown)


ryanking13 (approved):
Sounds all valid.

PR #379, opened 2026-06-12
refactor: remove dead code and apply Python 3.11+ modernizations

henryiii:

Part of #376.

This PR applies several verified code simplifications and modernizations to the `pyodide-build` codebase:

1. **Remove dead `_remove_path`** (`recipe/cleanup.py`): This helper was defined but never called anywhere in the repo. Cleanup is handled exclusively by `RecipeBuilder.clean()`.

2. **Simplify `_create_symlink_dir`** (`pypabuild.py`): Dropped the unused `env: dict[str, str]` parameter and converted the function from a `@contextmanager` (noted with a TODO comment) to a plain function returning the symlink directory path.

3. **Use `contextlib.chdir`** (`common.py`): Replaced the hand-rolled `chdir` context manager with a re-export of `contextlib.chdir` (available since Python 3.11).

4. **Use `hashlib.file_digest`** (`common.py`): Replaced the manual chunked SHA-256 loop in `_get_sha256_checksum` with `hashlib.file_digest(fd, "sha256").hexdigest()` (available since Python 3.11).

5. **Fix latent dataclass eq/hash break** (`recipe/graph_builder.py`): `PythonPackage`, `SharedLibrary`, and `StaticLibrary` were decorated with plain `@dataclasses.dataclass`, which regenerates `__eq__` and silently sets `__hash__ = None`, overriding `BasePackage`'s hand-written `__hash__`/`__eq__`. Changed to `@dataclasses.dataclass(eq=False)`. Added a test to verify subclass instances are hashable and use name/version equality.

6. **Deduplicate compression logic** (`common.py`, `_py_compile.py`): Extracted a `_zip_compression(level)` helper to remove the identical `if compression_level > 0: ZIP_DEFLATED else ZIP_STORED` block that appeared in three places.

7. **Update stale comment** (`build_env.py`): The comment said the manual `cpXY-none-any` Tag yield could be removed "once packaging 22.0" — but `cpython_tags` still does not emit this tag in packaging 26.x, so it is still required. Updated the comment to reflect this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,31 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- Removed dead `_remove_path` helper from `recipe/cleanup.py` (never called; cleanup
+  delegates entirely to `RecipeBuilder.clean()`).
+- Simplified `_create_symlink_dir` in `pypabuild.py`: dropped the unused `env`
+  parameter and converted it from a context manager to a plain function.
+- Replaced the hand-rolled `chdir` context manager in `common.py` with a re-export
+  of `contextlib.chdir` (Python 3.11+).
+- Replaced the manual chunked SHA-256 loop in `common._get_sha256_checksum` with
+  `hashlib.file_digest` (Python 3.11+).
+- Made `PythonPackage`, `SharedLibrary`, and `StaticLibrary` use
+  `@dataclasses.dataclass(eq=False)` so they inherit `BasePackage`'s hand-written
+  `__hash__`/`__eq__` rather than having `__hash__` silently set to `None` by the
+  generated `__eq__`.
+- Extracted `_zip_compression` helper in `common.py` to deduplicate the
+  `compression_level > 0 → ZIP_DEFLATED/ZIP_STORED` logic shared by
+  `make_zip_archive`, `repack_zip_archive`, and `_py_compile._compile`.
+- Updated a stale comment in `build_env.py` about the manual `cpXY-none-any` Tag:
+  `packaging.tags.cpython_tags` still does not emit this tag as of packaging 26.x,
+  so the manual yield is still required.
+
+Part of [#376](https://github.com/pyodide/pyodide-build/issues/376).
+
 ## [0.35.1] - 2026/06/13
 
 ### Fixed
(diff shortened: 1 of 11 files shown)


pyodide_build/common.py
```diff
@@ -333,14 +333,7 @@ def find_missing_executables(executables: list[str]) -> list[str]:
...
 
-@contextmanager
-def chdir(new_dir: Path) -> Generator[None, None, None]:
-    orig_dir = Path.cwd()
-    try:
-        os.chdir(new_dir)
-        yield
-    finally:
-        os.chdir(orig_dir)
+chdir = contextlib.chdir
```
hoodmane: We should replace `from common import chdir` with `from contextlib import chdir` at point of use.

CHANGELOG.md
```diff
@@ -5,6 +5,32 @@ All notable changes to this project will be documented in this file.
...
+- Made `PythonPackage`, `SharedLibrary`, and `StaticLibrary` use
+  `@dataclasses.dataclass(eq=False)` so they inherit `BasePackage`'s hand-written
+  `__hash__`/`__eq__` rather than having `__hash__` silently set to `None` by the
+  generated `__eq__`.
+- Extracted `_zip_compression` helper in `common.py` to deduplicate the
+  `compression_level > 0 → ZIP_DEFLATED/ZIP_STORED` logic shared by
+  `make_zip_archive`, `repack_zip_archive`, and `_py_compile._compile`.
+- Updated a stale comment in `build_env.py` about the manual `cpXY-none-any` Tag:
+  `packaging.tags.cpython_tags` still does not emit this tag as of packaging 26.x,
+  so the manual yield is still required.
```
hoodmane: Are any of these user-facing changes?

henryiii: I don't think so

pyodide_build/common.py
henryiii: :robot: _AI text below_ :robot:

Done in 9b0b87f — the re-export is removed and `builder.py`, `unvendor.py`, and `test_cli_xbuildenv.py` now use `from contextlib import chdir` directly.

CHANGELOG.md
henryiii: :robot: _AI text below_ :robot:

Agreed — none are user-facing. The CHANGELOG entry is reverted in 9b0b87f.

pyodide_build/build_env.py
```diff
@@ -277,7 +277,7 @@ def pyodide_tags_() -> Iterator[Tag]:
     python_version = (int(PYMAJOR), int(PYMINOR))
     yield from cpython_tags(platforms=PLATFORMS, python_version=python_version)
     yield from compatible_tags(platforms=PLATFORMS, python_version=python_version)
-    # Following line can be removed once packaging 22.0 is released and we update to it.
+    # packaging's cpython_tags does not emit cpXY-none-any (checked against packaging 26.x); keep this manual tag.
```
hoodmane:
```suggestion
    # packaging's cpython_tags does not emit cpXY-none-any
```

hoodmane (approved):
Thanks!

PR #380, opened 2026-06-12
Cache wheel metadata and simple-index lookups in the PyPI resolver

henryiii:

Part of #376

## Problem

The out-of-tree PyPI dependency resolver (`pyodide build --build-dependencies` / `-r requirements.txt`, in `pyodide_build/out_of_tree/pypi.py`) does redundant network work during resolution:

- **Whole wheels downloaded just to read `METADATA`.** `get_metadata_for_wheel` did `requests.get(url).content` on the full wheel (potentially tens of MB) to extract a single `METADATA` file, with no caching by URL. Because `Candidate` objects are recreated on every `find_matches` call, the per-instance `_metadata` attribute never helped across resolution rounds/backtracking, so the same wheel could be downloaded repeatedly.
- **Simple index re-fetched and `PackageFinder` rebuilt on every `find_matches` call.** `get_project_from_pypi` constructed a brand-new `unearth.PackageFinder` (new HTTP session, no connection reuse) and re-fetched the package's index page every time resolvelib re-evaluated an identifier — which happens whenever a new requirement is merged and on every backtrack.

## Fix

- `get_metadata_for_wheel` now tries **PEP 658** first: PyPI serves the metadata at `<wheel-url>.metadata` (a few KB), and it falls back to the full-wheel download when that 404s (other indexes may not support it). The parsed metadata is cached per URL via `functools.lru_cache`.
- Candidate discovery for a project is cached by name through an `lru_cache`d helper returning raw `(name, version, url)` tuples — candidates for a project don't depend on extras, only the `Candidate` objects built from them do — and a single shared `PackageFinder`/HTTP session is reused across the whole resolution.
- All three new caches are registered in the `reset_cache` test fixture so tests stay isolated.

These changes are caching layers only; `find_matches`' extras loop and `download_or_build_wheel`'s signature are untouched.

## Tests

Extended `pyodide_build/tests/test_pypi.py` (using `pytest-httpserver` with request counting):

- metadata is fetched via the `.metadata` URL when available, without downloading the wheel, and is cached (server hit once);
- falls back to the full wheel when `.metadata` 404s, then caches it;
- the simple index is fetched only once per project across repeated `get_project_from_pypi` calls with different extras.

The existing `test_pypi.py` suite and the full unit suite (`pytest pyodide_build -m "not integration"`) pass (pre-existing, unrelated failures in `test_cli`/`test_xbuildenv` aside).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   candidates for a project are cached by name and the `PackageFinder`/HTTP
   session is shared, and wheel metadata is now read via PEP 658
   (`<wheel-url>.metadata`, falling back to the full wheel) and cached per URL.
-  [#376](https://github.com/pyodide/pyodide-build/issues/376)
+  [#376](https://github.com/pyodide/pyodide-build/issues/376),
+  [#380](https://github.com/pyodide/pyodide-build/pull/380)
 
 ## [0.35.1] - 2026/06/13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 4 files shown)


ryanking13:
We are actually planning to remove this feature soon in #362. So let me close this. Thanks anyways!

PR #381, opened 2026-06-12
fix: correct four small bugs in wheel matching, clean, elapsed time, and config loading

henryiii:

Part of #376

## Summary

This PR fixes four verified bugs that were each small in scope but each causing
real incorrect behaviour in production:

**Bug 1 — `continue` should be `break` in `_find_matching_wheels`
(`pyodide_build/common.py`)**

A wheel whose filename contains a compressed tag set (e.g.
`pyodide_2024_0_wasm32.pyemscripten_2024_0_wasm32`) is expanded by
`packaging` into multiple `Tag` objects. The inner loop iterated over all
`supported_tags` and used `continue` after a match, causing the same wheel
path to be yielded once per matching tag. `find_matching_wheel` then raised
`RuntimeError("Found multiple matching wheels")` listing the same file twice.
Fixed by changing `continue` to `break` so each wheel is yielded at most once.
Also corrected the implicit-Optional annotation `version: str = None` →
`version: str | None = None` on `find_matching_wheel`.

**Bug 2 — `pyodide clean recipes` silently cleans "always"-tagged packages
(`pyodide_build/recipe/cleanup.py`)**

`resolve_targets` called `loader.load_recipes` without `load_always_tag=False`,
so any explicit request (e.g. `pyodide clean recipes mypackage`) would also
clean every package tagged `always`. Fixed by passing `load_always_tag=False`.

**Bug 3 — Elapsed-time log drops the hours component for builds >= 1 h
(`pyodide_build/recipe/graph_builder.py`)**

`PackageStatus.finish` used `datetime.fromtimestamp(elapsed_time, tz=UTC)` to
format a duration. For a 3700 s build this displayed "1m 40s" instead of "1h
1m 40s". Replaced with straightforward integer `divmod` arithmetic. The unused
`import datetime` was removed.

**Bug 4 — Non-string values in `[tool.pyodide.build]` crash with an opaque
`TypeError` (`pyodide_build/config.py`)**

`_load_config_file` passed raw TOML values to `_environment_substitute_str`
which calls `re.sub(..., string)`. A natural TOML spelling such as
`skip_emscripten_version_check = true` produced `TypeError: expected string or
bytes-like object, got 'bool'` with no indication of the offending key. Fixed
by converting TOML booleans to `"1"` (true) or `"0"` (false) — matching the
env-var convention where `SKIP_EMSCRIPTEN_VERSION_CHECK` defaults to `"0"` —
and casting other non-string scalars with `str()`.

## Test plan

- New test `test_find_matching_wheels_no_duplicate_yield` verifies that a wheel
  with a compressed tag set matching two `supported_tags` entries is yielded
  exactly once by `_find_matching_wheels`.
- New test `test_find_matching_wheel_compressed_tags_no_runtime_error` verifies
  `find_matching_wheel` does not raise for such a wheel.
- New tests `test_resolve_targets_does_not_include_always_tagged_packages` and
  `test_clean_recipes_does_not_clean_always_tagged_packages` verify the
  always-tag exclusion.
- New parametrised test
  `test_package_status_elapsed_time_formatting` covers 0 s, sub-minute,
  sub-hour, and multi-hour cases.
- New test `test_load_config_from_file_non_string_values` exercises TOML
  `true`/`false` values in `[tool.pyodide.build]`.
- Full unit suite (`pytest pyodide_build -m "not integration"`): 337 passed,
  1 skipped, 1 xfailed, 4 pre-existing failures (unrelated to this PR).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,37 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- Fixed `_find_matching_wheels` yielding the same wheel multiple times when its
+  compressed tag set (e.g. `pyodide_2024_0_wasm32.pyemscripten_2024_0_wasm32`)
+  matches more than one entry in `supported_tags`, which caused
+  `find_matching_wheel` to raise `RuntimeError("Found multiple matching
+  wheels")` for a single file. Also corrected the implicit-Optional annotation
+  `version: str = None` to `version: str | None = None` in `find_matching_wheel`.
+  Part of [#376](https://github.com/pyodide/pyodide-build/issues/376)
+
+- Fixed `pyodide clean recipes` incorrectly cleaning packages tagged `always`
+  even when those packages were not in the requested target list. The fix passes
+  `load_always_tag=False` to `loader.load_recipes` inside `resolve_targets`.
+  Part of [#376](https://github.com/pyodide/pyodide-build/issues/376)
+
+- Fixed elapsed-time log messages dropping the hours component for builds
+  lasting ≥ 1 hour (e.g. 3700 s was shown as "1m 40s" instead of "1h 1m 40s").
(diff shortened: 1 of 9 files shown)


ryanking13 (approved):
All makes sense to me. Let me tidy up the change log a bit and merge

agriyakhetarpal (approved):

PR #382, opened 2026-06-12
Sanitize Content-Disposition filename to prevent path traversal

henryiii:

Part of #376

## Summary

- `_extract_tarballname` in `pyodide_build/recipe/builder.py` trusted the `Content-Disposition: filename=...` header verbatim. A malicious or compromised mirror could supply a value like `../../evil.tar.gz` or `/abs/path.tar.gz`, causing `self.build_dir / tarballname` to resolve outside the build directory.
- The fix introduces `_sanitize_filename()` which strips all Unix and Windows directory separators by taking only the final path component via `PurePosixPath.name` + `PureWindowsPath.name`, and rejects empty/`.`/`..` values with a fallback to the URL-derived filename.
- Member-level tar/zip extraction was already protected (tar `"data"` filter, `_unpack_zipfile`); this closes the remaining gap.

## Tests

Five parametrized cases covering Unix path traversal, absolute paths, empty filenames, Windows backslash traversal, and the normal happy path. Full unit suite passes (4 pre-existing failures on `main`, unrelated).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   to prevent path traversal: the extracted filename is now reduced to its final path
   component only (stripping any Unix or Windows directory separators), so a malicious
   mirror cannot write files outside the build directory.
-  [#376](https://github.com/pyodide/pyodide-build/issues/376)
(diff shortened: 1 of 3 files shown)


ryanking13 (approved):
Makes sense to me.

PR #383, opened 2026-06-12
Fix bash runner hang on early shell exit and cached env mutation

henryiii:

Part of #376

## Summary

Two fixes in `pyodide_build/recipe/bash_runner.py`:

- **Build hang on `exit 0`:** `run_unchecked` appends an env-dump command after the user script. If the script terminated the shell itself (`exit 0`, or `exec` of a program that exits 0), bash returned rc 0 without the env dump ever running, and the parent blocked forever on `readline()` — no EOF arrived because the parent still held the pipe's write end. Any recipe `build/script` or `build/post` ending in `exit 0` deadlocked the build. Now a fresh pipe is created per call and the parent closes its copy of the write end before reading, so an early shell exit yields EOF; in that case the previous environment is kept (with a debug log).
- **Cached env dict mutation:** `get_bash_runner` called `env.update(extra_envs)` directly on the dict returned by the `@functools.cache`d `get_build_environment_vars()`, permanently injecting per-package vars (`PKGDIR`, `PKG_VERSION`, `DISTDIR`, plus `os.environ`) into the cached value — later packages built in the same process saw the previous package's values. The dict is now copied first.

## Tests

- `test_run_unchecked_exit_zero_does_not_hang` (thread watchdog so a regression fails rather than hangs; verifies env stays usable and a later script still updates it)
- `test_run_unchecked_does_not_leak_fds`
- `test_get_bash_runner_does_not_mutate_cached_env`

Full unit suite passes (4 pre-existing failures on `main`, unrelated).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- Fixed a hang when a recipe `build/script` or `build/post` ends by exiting the
+  shell itself (e.g. `exit 0`). The build no longer blocks forever waiting for an
+  environment dump that never runs; the previous environment is kept instead.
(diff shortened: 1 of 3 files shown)


ryanking13 (approved):
Makes sense to me. Let's see if the integration tests pass

PR #384, opened 2026-06-12
Fix skeleton update format fallback and sdist URL generation

henryiii:

Part of #376

## Bug 1 — dead-code format fallback in `update_package`

`source_fmt = source_fmt or old_fmt` made `source_fmt` always truthy, so the fallback branches `["wheel","sdist"]` / `["sdist","wheel"]` were unreachable. A recipe using an sdist whose new release ships only wheels raised `MkpkgFailedException` instead of transparently updating to the wheel (and vice versa). Format resolution now happens after the version check and restores the intended semantics: explicit `source_fmt` is strict, absent `source_fmt` prefers the old format with fallback to the other.

## Bug 2 — sdist predictable URL ignores actual filename

The predictable-URL generation always produced `{normalized_name}-{version}.tar.gz` for sdists, ignoring the actual filename from PyPI metadata. `.zip` sdists and packages with non-normalized filenames (e.g. `ruamel.yaml`) got URLs that 404 while the stored `sha256` is from the real file, producing unbuildable recipes. The actual `filename` from the PyPI metadata is now used.

## Tests

Seven new tests: sdist→wheel and wheel→sdist fallback, strict explicit `source_fmt`, three parametrized predictable-URL cases (`.tar.gz`, `.zip`, non-normalized names), and an end-to-end `.zip` sdist URL check. Full unit suite passes (4 pre-existing failures on `main`, unrelated).


agriyakhetarpal (approved):
Thanks! A note: a predictable URL of the form https://files.pythonhosted.org/packages/source/s/scikit_learn/scikit_learn-1.6.1.tar.gz also works right now because PyPI normalises it before downloading, but I guess it makes a case for normalisation, which is a good thing. The https://files.pythonhosted.org/packages/source/r/ruamel_yaml/ruamel.yaml-0.18.6.tar.gz change is useful to have – I don't think we've run into that one before.

Let me make the CHANGELOG less verbose while fixing the conflicts and merge.

agriyakhetarpal:
@henryiii, just to confirm: `zip`-style sdists are all pre-PEP-517 behaviour, right? I don't think (or hope) anyone is doing something like `python setup.py sdist --formats=zip` anymore, but I am fine with keeping this just in case.

PR #386, opened 2026-06-12
perf: stream downloads, lazy xbuildenv path, skip pure-wheel repack

henryiii:

## Summary

Three small performance improvements to the pyodide-build toolchain. Part of #376.

- **Stream archive downloads**: `download_and_unpack_archive` in `common.py` previously called `resp.read()` which buffers the entire archive (potentially hundreds of MB for xbuildenv/emsdk tarballs) in RAM before writing to disk. Changed to stream directly with `shutil.copyfileobj(resp, f)`, reducing peak memory usage proportionally to archive size.

- **Lazy xbuildenv default path resolution**: `DEFAULT_PATH = default_xbuildenv_path()` in `cli/build.py` and `cli/xbuildenv.py` ran at module import time — triggering a `ConfigManager` build, a parent-directory `pyproject.toml` walk, write-access probes, and a potential `RuntimeError` — on *every* `pyodide` invocation for *every* subcommand. Changed both modules to `default=None` on all `--path`/`--xbuildenv-path` options and added an `if path is None: path = default_xbuildenv_path()` guard inside each command body so the cost is deferred until it is actually needed.

- **Skip wheel unpack/repack for pure-Python wheels**: `out_of_tree/build.py` previously called `modify_wheel` (two `python -m wheel` subprocess round-trips + full unzip/rezip) unconditionally, even for pure-Python wheels where `replace_so_abi_tags` does nothing. Now checks the wheel's zipfile namelist for any `.so` entries before deciding whether to unpack. The logic is extracted into a helper `_check_and_maybe_modify_wheel` for testability.

## Test plan

- [x] `pyodide_build/tests/test_xbuildenv.py::TestCrossBuildEnvManager::test_download` — exercises `download_and_unpack_archive` against `pytest-httpserver`; still passes with streaming.
- [x] `pyodide_build/tests/test_cli.py::test_cli_import_does_not_resolve_xbuildenv_path` — new test that reloads `cli.build` and `cli.xbuildenv` under a monkeypatched `default_xbuildenv_path` and asserts it was never called.
- [x] `pyodide_build/tests/test_oot_build.py::TestSoDetectionSkip` — new tests: pure wheel → `replace_so_abi_tags` not called; wheel with `.so` → called.
- [x] Full unit suite (`pytest pyodide_build -m "not integration"`) passes (320 passed, 1 skipped, 1 xfailed; pre-existing failures unchanged).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,26 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- Archive downloads (xbuildenv, emsdk) now stream directly to a temporary file
+  instead of buffering the entire archive in RAM, reducing peak memory usage for
+  large downloads. Part of [#376](https://github.com/pyodide/pyodide-build/issues/376).
+
+- The default xbuildenv path is now resolved lazily — only when a CLI command
+  body actually needs it — instead of at module import time. This avoids
+  triggering a `ConfigManager` build, filesystem probes, and a potential
+  `RuntimeError` on every `pyodide` invocation, regardless of the subcommand
+  used. Part of [#376](https://github.com/pyodide/pyodide-build/issues/376).
+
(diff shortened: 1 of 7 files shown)


ryanking13 (approved):

ryanking13:
Cleaned up changelog as these are not user-visible changes.

PR #387, opened 2026-06-15
Support π and 𝜋 in Pyodide venv launchers for Pyodide 314.x xbuildenvs

agriyakhetarpal:
> [!TIP]
> Moved to #388

### DIFF
--- a/pyodide_build/out_of_tree/venv.py
+++ b/pyodide_build/out_of_tree/venv.py
@@ -37,6 +37,16 @@ SUPPORTED_VIRTUALENV_OPTIONS = [
     "--no-periodic-update",
 ]
 
+# See https://github.com/python/cpython/issues/119535
+# fmt: off
+PI_ALIASES = (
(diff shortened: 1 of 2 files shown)


agriyakhetarpal:
Wow, this really broke because of the branch name! `python-π,𝜋-fun` breaks our test suite because GHA exposes its non-ASCII bits via env vars like `GITHUB_HEAD_REF`/`GITHUB_REF`/`GITHUB_REF_NAME`/etc... On the macOS runner, that apparently gets mangled into the environment with a malformed UTF-8 sequence.

And each of the 66 failing tests goes through `_get_make_environment_vars`, which dumps the entire process environment (including `GITHUB_HEAD_REF`) to stdout. `run_command` then does `subprocess.run()` which strictly decodes that output as UTF-8 and then breaks.

agriyakhetarpal:
I renamed the branch to `python-pi-fun` and will recreate the PR with it.

PR #396, opened 2026-07-08
pip patch: Update packaging.tags patch to handle newer versions

hoodmane:
packaging 26.2 already contains support for Emscripten, we just have to patch in the legacy pyodide_xxx_wasm32 tags. Older versions still need the full patch.


agriyakhetarpal (approved):
Makes sense to me, thanks @hoodmane!

PR #397, opened 2026-07-08
Factor out pip wrapper into its own file

hoodmane:


pyodide_build/out_of_tree/venv.py
```diff
@@ -548,22 +393,10 @@ def _create_pip_script(self) -> None:
...
-                        sys.argv[0] = sys.argv[0].replace('{pip_wrapper_name}', 'pip')
-                        sys.exit(main())
-                    """
-                )
-            ).replace("\\", "\\\\")  # Escape backslashes for Windows batch files
-        )
+        pip_wrapper_src = Path(__file__).parent / "pip_wrapper.py"
+        shutil.copy(pip_wrapper_src, self.pip_wrapper_path)
+        with (self.venv_bin / "pyodide_pip_config.json").open("w") as f:
+            json.dump(self._get_pyodide_pip_config(), f)
```
ryanking13: I think you can use `Path.write_text` here and remove the context manager.

pyodide_build/out_of_tree/pip_wrapper.py
```diff
@@ -0,0 +1,193 @@
+import json
```
ryanking13, line 3: Let's add a comment that this file is being copied to the build directory so it should not import any pyodide-build source files.

pyodide_build/out_of_tree/pip_wrapper.py
```diff
@@ -0,0 +1,193 @@
...
+    exe_suffix: str
+    pip_patched_name: str
+    pip_wrapper_name: str
+    platform_data: tuple[str, str, str, str, str]
+    pyodide_platform: str
+    sysconfigdata_dir: str
+
+
+with (Path(__file__).parent / "pyodide_pip_config.json").open("r") as _config_file:
+    CONFIG = Config(**json.load(_config_file))
```
ryanking13: same, the read_text (or read_byte) can work here.

ryanking13 (approved):
I didn't read the factored out `pip_wrapper.py` but I believe it was a mechanical refactor except for passing the configs. Otherwise, I left some minor comments.

PR #400, opened 2026-07-14
Fix mypy

hoodmane:
When pyodide_build was unvendored from pyodide, the mypy config was broken. This fixes the mypy config and all the type check failures that have been introduced since pyodide_build was unvendored.


pyodide_build/recipe/bash_runner.py
```diff
@@ -42,6 +42,7 @@ def run_unchecked(self, cmd: str, **opts: Any) -> subprocess.CompletedProcess[st
         # block forever, since it would still hold the write end open and never
         # see EOF. See https://github.com/pyodide/pyodide-build/issues/376.
         fd_read, fd_write = os.pipe()
+        fd_write_open = True
```
hoodmane: This isn't a good change I'm going to put it back I think.

ryanking13 (approved):

PR #401, opened 2026-07-19
Deploy unvendored tests in a zipfile format

ryanking13:
This makes https://github.com/pyodide/pyodide/pull/6354 easier, as we don't need to implement our own tar decompressor


hoodmane (approved):
Thanks!

PR #402, opened 2026-07-20
Replace pydantic with attrs+cattrs

hoodmane:


pyproject.toml
```diff
@@ -103,6 +104,8 @@ python_version = "3.12"
 show_error_codes = true
 warn_unreachable = true
 enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
+# pyodide-lock is a pydantic-based dependency; the plugin type-checks
+# construction of its models (pyodide-build's own models use attrs/cattrs).
```
hoodmane: TODO: first release pyodide-lock with no pydantic, then remove this.

pyodide_build/xbuildenv_releases.py
```diff
@@ -199,6 +204,19 @@ def get_release(
...
+_converter = cattrs.Converter(detailed_validation=False)
+# CrossBuildEnvReleaseSpec uses extra="ignore" (default), CrossBuildEnvMetaSpec
+# uses extra="forbid".
+_converter.register_structure_hook(
+    CrossBuildEnvMetaSpec,
+    make_dict_structure_fn(
+        CrossBuildEnvMetaSpec,
+        _converter,
+        _cattrs_forbid_extra_keys=True,
+    ),
```
ryanking13, line 229: Maybe let's allow extras for future spec changes. I think it was a mistake (by me) to disallow extra fields.

pyproject.toml
ryanking13: yeah, please do

ryanking13 (approved):
Thanks! I left a minor comment otherwise looks good

pyodide_build/xbuildenv_releases.py
hoodmane: Let's do it in a followup PR so that we have one commit that rewrites and a second commit that makes changes.

ryanking13: Sure, sounds good

PR #404, opened 2026-07-21
Apply RUSTFLAGS only to Emscripten target

hoodmane:


pyodide_build/config.py
```diff
@@ -251,7 +251,7 @@ def _parse_makefile_envs(
     "pythoninclude": "PYTHONINCLUDE",
     "pyversion": "PYVERSION",
     "cpythoninstall": "CPYTHONINSTALL",
-    "rustflags": "RUSTFLAGS",
+    "rustflags": "CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_RUSTFLAGS",
```
ryanking13, line 259: This change means that the `RUSTFLAGS` coming from the `Makefile.envs` will not be applied by default, which is kinda regression. It is an expected behavior?

hoodmane: No it's not intended. Can we copy from RUSTFLAGS into this other variable?

ryanking13: You need to patch this function a little bit:

https://github.com/pyodide/pyodide-build/blob/ead27015e924366493192984f525966d45e0090a/pyodide_build/config.py#L164-L173

I guess you what you need to do is something like

```diff
                varname = line[0:equalPos]
+              if varname == "RUSTFLAGS":
+                   # rename RUSTFLAGS to CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_RUSTFLAGS
+                   # so that it is only applied to WASM32 target.
+                   # Makefile.envs of older Pyodide versions have RUSTFLAGS not CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_RUSTFLAGS
+                    varname = "CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_RUSTFLAGS"
                if varname not in BUILD_VAR_TO_KEY:
                    continue
```

hoodmane: Thanks!

agriyakhetarpal (approved):
Makes sense to me, thanks!

hoodmane:
Thanks for the review!

PR #406, opened 2026-07-25
Allow extra fields in xbuildenv_releases.py

hoodmane:
As suggested by @ryanking13


agriyakhetarpal (approved):
LGTM as it is in line with the changes made for #349. Thanks @hoodmane!

PR #408, opened 2026-07-26
Test both Python 3.13 and 3.14, i.e., against the Pyodide 0.29.x and 314.x release lines

agriyakhetarpal:
We have only ever tested against one Pyodide version at a time, but I've been thinking that we should expand this a little bit. Looks like this doesn't add too much to the CI time, which is good.

This PR adds `python-version` as a matrix dimension across the workflow.
- `integration-test` goes from 9 to 17 jobs. The `include` entries are duplicated per Python, apart from the `stable-debug` one, which stays 3.14-only because we didn't have stable-debug environments in 3.13.
- `test` jobs go from 2 to 4, `test-venv` from 3 to 6, and `windows-integration-test` from 1 to 2.

There is now one dummy xbuildenv fixture per host Python. Ths means we now have `xbuildenv-test-3.13.tar.gz` and `xbuildenv-test-3.14.tar.gz`, and `dummy_xbuildenv_url` picks the one matching whatever Python version is running the tests. I added a script (mostly AI-generated but I verified it and made it less verbose) to make it easier to build the dummy xbuildenvs. @ryanking13, I'm curious in case you had one like this that you were using, or if we were always doing this by hand :D

My main reason for this PR is because both versions' ABIs are capable of publishing pyemscripten wheels, so we should commit to higher standards of testing for these versions in pyodide-build.

The minimum-pyodide testing we established via #149 does not use Python 3.14, because we only have four 314.x releases and 314.0.0 is pretty recent, so it is much less important for anything that is not the oldest Python version we support. I think it might also be the right time to set the minimum Python version to 3.13, as 3.12 (Pyodide 0.26) has no Rust configuration in `Makefile.envs`.


agriyakhetarpal:
Argh, orjson is failing to build on 3.14. We never tested on Python 3.14, so we're only discovering this today ...

Edit: resolved by finding an orjson version that builds on Pyodide 0.28, 0.29, and 314.0 with an appropriate MSRV.

ryanking13:
> to make it easier to build the dummy xbuildenvs. @ryanking13, I'm curious in case you had one like this that you were using, or if we were always doing this by hand :D

I was doing it by my hand, not hoping to do it often 😂. Thanks for adding it.

ryanking13 (approved):
Thanks!

agriyakhetarpal:
Thanks for the quick review @ryanking13!

PR #412, opened 2026-07-26
Add build_extras requirements field to recipe to allow adding extra build-requires

hoodmane:


agriyakhetarpal:
Without having looked at this in detail, is this something we require on a frequent basis when building recipes, since I assume we can always add build-time requirements by patching the packages' `pyproject.toml` or `setup.py` files? It looks like it's coming more out of a need for convenience – something which I am not opposed to; I just want to know how many packages in our distribution require this.

hoodmane:
Currently just scipy.

> we can always add build-time requirements by patching the packages' pyproject.toml or setup.py files?

Yes but this is much more annoying.

hoodmane:
We need often to pin or cap build requirements or transitive requirements for packages. Sometimes we need to substitute a fork of a build requirement. All these things can be done much more conveniently with this field than with a patch.

hoodmane:
Especially in the common case where e.g., a setuptools release breaks a bunch of our packages and we want to add a cap everywhere. Adding one patch per recipe is frustrating.

hoodmane:
Using it here to substitute a fork of gast that supports Python 3.15:
https://github.com/pyodide/pyodide/pull/6355/changes#diff-d6d5fb2d234d6c6bafa1cff3848a6f49e87a62111e50baab21b4e954ae2b85beR154-R159

agriyakhetarpal:
Agreed, the use case makes sense, and I am aligned with it. But we should note the way this is currently implemented:

> Sometimes we need to substitute a fork of a build requirement. All these things can be done much more conveniently with this field than with a patch.

This is additive, and doesn't replace existing build requirements. How do we make it replace the build requirements entirely? Perhaps another field `override-requirements` where we completely ignore the build requirements coming from the package, and add our own?

hoodmane:
> This is additive, and doesn't replace existing build requirements.

It seems to work fine? If it doesn't work in the future, then we can modify it. But you will note that pyodide/pyodide#6355 is using this change and CI is green.

agriyakhetarpal:
Ah, your new comment came up just as I was typing mine. I would suggest making this clearer in the recipe format docs. pyodide/pyodide#6355 looks good to me, then.

agriyakhetarpal:
i.e., what I mean is that `requirements.build` from a bird's eye view seems to me as if it overrides everything but it's actually adding extra requirements. Could we rename the field to something better...

hoodmane:
How about `build_extras`

agriyakhetarpal:
Works for me!

pyodide_build/recipe/spec.py
```diff
@@ -155,6 +155,7 @@ def _check_config(self) -> None:
 class _RequirementsSpec:
     run: list[str] = _afield(factory=list)
     host: list[str] = _afield(factory=list)
+    build_extras: list[str] = _afield(factory=list)
```
agriyakhetarpal: This should have an `alias="build-extras"`, since no other field currently has underscores, so this breaks our conventions. We've always had hyphens.


```suggestion
    build_extras: list[str] = _afield(factory=list, alias="build-extras")
```

pyodide_build/tests/test_pypabuild.py
```diff
@@ -387,3 +394,90 @@ def _fake_isolated(
...
+        str(tmp_path / "dist"),
+        "wheel",
+        {},
+        extra_build_requires=["cython", "pkgconfig"],
+    )
+
+    assert wheel.endswith("pkg-1.0-py3-none-any.whl")
+    # The build system requires and the extra build requires are installed
+    # together, before the backend is queried for its dynamic requirements.
+    assert installed == [{"setuptools", "cython", "pkgconfig"}, {"wheel"}]
```
agriyakhetarpal: What should we do when a backend is queried for its dynamic requirements, and a conflict arises between what we get from it and what we install as part of the extra build requirements juxtaposition?

Imagine a package recipe that mentions Ninja 1.2. We add an extra build requirement through the field so that we pull in Ninja 1.4. Suppose the PEP 517 hook (`get_requires_for_build_wheel`) is then called and that dynamically returns Ninja 1.3. That would mean we don't actually get Ninja 1.4 like we wanted to have, through the override, right? i.e., we don't want to be clobbering `build-extras` when the PEP 517 hook returns an incompatible pin.

pyodide_build/pypabuild.py
agriyakhetarpal: I guess we should add your change from above here as well (see my previous comment)?

```suggestion
        install_reqs(build_env, env, build_reqs | set(extra_build_requires))
```

(Not sure if the comment renders properly in the GitHub UI, but essentially my ask is to mirror your change from lines 309-311 to line 335 as well).

agriyakhetarpal (commented):
This reads better to me now. I have a few questions and suggestions, but overall LGTM.

Could you please add:
- an integration recipe test that exercises this
- a unit test that adds extra build requirements with PEP 508 markers (I checked locally that these already work, but we've already had a regression for this before, so one test would actually be great)
- a CHANGELOG entry?

Thanks @hoodmane!

P.S. I was going to suggest that we should add a note about this new field to the "The meta.yaml specification" page but I couldn't find it. As it turns out, it's gone with https://github.com/pyodide/pyodide/pull/6181 and we never added that page to pyodide-build's docs. We should add it at some point.

ryanking13 (approved):
Thanks, yeah this is often annoying when we need to deal with some old packages or when there is a breaking changes in the build system.

Since we are slowly deprecating the recipes system (at least in my opinion), I am okay with adding some adhoc features to the recipes to make our life easier. People who are building packages out of tree would rarely use recipes anyways.

hoodmane:
I still think the recipes play an important role in testing new emscripten versions and new abis and finding Emscripten bugs and other problems. If we reduce the number too much, the risk is that problems are discovered downstream much later by people without the resources to fix them.

Especially since maintaining the recipes is quite a lot easier with langage models, so the cost has gone down.

pyodide_build/tests/test_pypabuild.py
hoodmane: We can update pyodide-build to handle this problem if it occurs.

PR #414, opened 2026-07-27
Enable xbuildenv source introspection via a marker file

agriyakhetarpal:
I don't particularly like the ergonomics of this PR... hence, this is ready for review, but not ready to merge just yet and I am scouting opinions.

The idea is to have a file called `.xbuildenv-source`, which will record where the xbuildenv came from – whether it's a release xbuildenv, a stable-debug xbuildenv, a nightly one, or a nightly-debug one. We can think of these as "channels". Of course, we also support custom URLs for the source of the xbuildenv, in which case we record the URL itself into the marker file.

LMK what you think about this!

### DIFF
--- a/pyodide_build/xbuildenv.py
+++ b/pyodide_build/xbuildenv.py
@@ -15,9 +15,12 @@ from pyodide_build.xbuildenv_releases import (
     CROSS_BUILD_ENV_METADATA_URLS,
     DEFAULT_SOURCE,
     CrossBuildEnvReleaseSpec,
+    ReleaseSource,
     SourceType,
+    SourceURL,
     cross_build_env_metadata_url,
     load_cross_build_env_metadata,
+    parse_source_url,
     source_for_metadata_url,
 )
 
@@ -37,7 +40,7 @@ class CrossBuildEnvManager:
         self,
         env_dir: str | Path,
         metadata_url: str | None = None,
-        source: SourceType | None = None,
+        source: ReleaseSource | None = None,
     ) -> None:
         """
         Parameters
@@ -64,7 +67,7 @@ class CrossBuildEnvManager:
         if source is None:
             source = source_for_metadata_url(self.metadata_url)
 
-        self.source: SourceType = source or DEFAULT_SOURCE
+        self.source: ReleaseSource = source or DEFAULT_SOURCE
 
         try:
             self.env_dir.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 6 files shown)


pyodide_build/xbuildenv.py
```diff
@@ -12,34 +12,62 @@
...
         self.env_dir = Path(env_dir).resolve()
-        self.metadata_url = metadata_url or cross_build_env_metadata_url()
+        self.metadata_url = metadata_url or cross_build_env_metadata_url(
+            source or DEFAULT_SOURCE
+        )
+        # If source is not provided, derive it from the metadata file, so that
+        # pointing PYODIDE_CROSS_BUILD_ENV_METADATA_URL at the nightly metadata
+        # can still record what was installed.
+        if source is None:
+            source = source_for_metadata_url(self.metadata_url)
```
ryanking13, line 68: I don't think this is needed. If a user passes any metadata URL, we should assume it is a custom URL.

pyodide_build/xbuildenv.py
```diff
@@ -72,6 +100,129 @@ def current_version(self) -> str | None:
...
+            "Reinstalling the cross-build environment at '%s': it %s, "
+            "but the '%s' source was requested.",
+            download_path,
+            reason,
+            self.source,
+        )
+        replaced_path = download_path.with_name(download_path.name + ".replaced")
+        shutil.rmtree(replaced_path, ignore_errors=True)
+        download_path.rename(replaced_path)
+        return replaced_path
```
ryanking13, line 194: Well, I don't like this behavior. How about adding the source to the xbuildenv directory instead when downloading? For example, we can use `0.29.3-debug` or `0.29.3-nightly` instead of sharing `0.29.3`. By doing that, we can remove the `SOURCE_MARKER_FILE` and use the directory name as the version name, which is consistent with the previous behavior.

pyodide_build/xbuildenv_releases.py
```diff
@@ -23,6 +24,56 @@
...
 
+# The published streams a cross-build environment can be installed from. These
+# are the ones with release metadata behind them, so only these can be asked for
+# by name, with --nightly and --debug.
+type ReleaseSource = Literal["stable", "stable-debug", "nightly", "nightly-debug"]
+
+# An environment installed with --url has no release behind it to name, so the
+# URL it came from is its source. A distinct type rather than a bare str, so
+# that a URL cannot be passed where a release stream is expected.
+SourceURL = NewType("SourceURL", str)
```
ryanking13, line 35: I don't quite get what the benefit of using a wrapped type is in this case. Of course, [parse don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/) is a good practice, but do we need any validation at all for this URL?

PR #416, opened 2026-07-27
Make cmake's try_compile use pywasmcross

hoodmane:
Testing this here:
https://github.com/pyodide/pyodide-recipes/pull/635


ryanking13:
Which package had issue with this? The change looks okay but I am curious.

hoodmane:
Three of them: igraph, healpy, and scikitlearn. But there's probably a lot more libraries out there that attempt to feature detect openmp.

ryanking13:
Sounds good, but I think this needs a changelog + version bump as this can be a breaking change for a few packages.

ryanking13 (approved):

PR #418, opened 2026-07-28
Add _cross-build-env-skip-install

hoodmane:
This adds a private field to the build spec which makes it not install the cross-build package into .artifacts. This can only be used for "leaf" xbuild packages for which we don't build any recipes that are dependents (e.g., scipy but not numpy). It is useful for pre-release Python versions when there is no native wheel available and we can't or don't want to build a native wheel.

Ideally, we would only install xbuild packages when they are used as a host dep of a recipe and that would have all the benefits of this automatically. I intend to do this in a followup and remove this key but that is a substantially larger diff.

### DIFF
--- a/pyodide_build/recipe/builder.py
+++ b/pyodide_build/recipe/builder.py
@@ -675,7 +675,10 @@ class RecipeBuilderPackage(RecipeBuilder):
                 Path(self.build_args.host_install_dir)
                 / f"lib/{python_dir}/site-packages"
             )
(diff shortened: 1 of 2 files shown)


ryanking13:
> This adds a private field to the build spec which makes it not install the cross-build package into .libs.

It doesn't matter, but I think we are not installing cross-build packages to `.libs`. I think we are installing them to `packages/.artifacts`.

ryanking13 (approved):

PR #419, opened 2026-07-28
pywasmcross: Disable openmp

hoodmane:
Emscripten 6.0.3 added support for openmp, and build systems started feature detecting it and using it. However, it requires pthreads and fails to link or to load. This updates pywasmcross to make it so that openmp is feature detected as not present.

### DIFF
--- a/pyodide_build/pywasmcross.py
+++ b/pyodide_build/pywasmcross.py
@@ -223,7 +223,7 @@ def replay_genargs_handle_linker_opts(arg: str) -> str | None:
         return None
 
 
-def replay_genargs_handle_argument(arg: str) -> str | None:
+def replay_genargs_handle_argument(arg: str) -> str | int | None:
     """
     Figure out how to replace a general argument.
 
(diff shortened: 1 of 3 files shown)


pyodide_build/tests/test_pywasmcross.py
```diff
@@ -287,3 +289,19 @@ def test_is_link_cmd():
     assert is_link_cmd(["test.so"])
     assert is_link_cmd(["test.so.1.2.3"])
     assert not is_link_cmd(["test", "test.a", "test.o", "test.c", "test.cpp", "test.h"])
+
+
+@pytest.mark.parametrize("flag", ["-fopenmp", "-fopenmp=libomp", "-fopenmp=libgomp"])
```
agriyakhetarpal, line 295: You could also add libiomp here

agriyakhetarpal (approved):
Thanks @hoodmane!

pyodide_build/pywasmcross.py
```diff
@@ -649,6 +659,8 @@ def handle_command(
         line = tmp
 
     new_args = handle_command_generate_args(line, build_args)
+    if isinstance(new_args, int):
+        return new_args
```
ryanking13: I am a little bit worried that people facing this would have difficulties debugging what has gone wrong. I wonder if we can show some useful error message in addtion to the non-zero return code.

hoodmane: I will update it to write something to standard out.

PR #421, opened 2026-08-03
Only install cross-build packages that are build dependencies

hoodmane:
This resolves a problem with building for Python 3.15, namely that scipy has no wheels on pypi yet but when building a package that depends on another xbuild package, say numpy, we try to install native scipy which tries to build native scipy which very likely fails.

Instead, only install the cross-build packages that appear as build dependencies.

### Example
```sh
git clone git@github.com:PyWavelets/pywt.git
cd pywt
uv python -p 315
uv pip install pyodide-build
.venv/bin/pyodide build .
```
#### Without this change

```
Creating isolated environment: venv+uv...
Using external uv from /home/hood/.local/bin/uv
Installing cross-build packages for 315.0.0a1   

RuntimeError: Failed to install cross-build packages: Using CPython 3.15.0b3 interpreter at: .venv/bin/python3
Resolved 4 packages in 5ms
   Building scipy==1.18.0
  × Failed to build `scipy==1.18.0`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `mesonpy.build_wheel` failed (exit status: 1)
```

#### With this change

Builds successfully.

### DIFF
--- a/pyodide_build/recipe/builder.py
+++ b/pyodide_build/recipe/builder.py
@@ -26,9 +26,12 @@ from pyodide_build.build_env import (
     _create_constraints_file,
     get_build_environment_vars,
     get_build_flag,
+    get_current_xbuildenv_manager,
     get_pyodide_root,
     get_pyversion_major,
     get_pyversion_minor,
+    get_unisolated_packages,
+    in_xbuildenv,
     pyodide_tags,
     replace_so_abi_tags,
     wheel_platform,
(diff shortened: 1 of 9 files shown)


pyodide_build/recipe/builder.py
```diff
@@ -310,12 +313,37 @@ def _build(self) -> None:
...
+
+        unisolated_packages = {name.lower() for name in get_unisolated_packages()}
+        needed = unisolated_packages & {
+            req.lower() for req in self.recipe.requirements.host
+        }
+
+        if needed:
+            get_current_xbuildenv_manager().ensure_cross_build_packages_installed(
+                needed
+            )
```
ryanking13:
```suggestion
        if not needed:
            return
        get_current_xbuildenv_manager().ensure_cross_build_packages_installed(
            needed
        )
```

pyodide_build/xbuildenv.py
```diff
@@ -83,43 +84,70 @@ def _path_for_version(self, version: str) -> Path:
...
+
+        host_site_packages = self._host_site_packages_dir(xbuildenv_pyodide_root)
+        installed = _installed_distributions(host_site_packages)
+
+        to_install = {}
+        for canonical in requested:
+            match = available.get(canonical)
+            if match is None:
+                # Not a cross-build package of this xbuildenv, nothing to do.
+                continue
```
ryanking13: It is weird that the callers of this function are already filtering the arguments by calling `get_unisolated_packages`, and then this function is checkign the packages again with `read_pinned_requirements`.

I think we can update this function to take the `(name, version)` tuple instead and remove this redundant check.

ryanking13 (approved):
I left some minor comments but otherwise looks good to me

PR #423, opened 2026-08-04
Fix CLI command names in docs

grden:
If you look at https://pyodide.org/en/stable/usage/api/pyodide-cli.html, most entries in the table of contents under `pyodide` are not showing the actual CLI commands but are instead repeated as `app` or `main`.

Reason is that Click infers the command name from the Python function name unless it is set explicitly. So `def main()` becomes `main`, and `def app()` becomes `app`, which does not match the pyodide.cli entry point names like `build`, `clean`.

This PR sets the Click command names explicitly, so sphinx correctly renders the command names.

& Same changes are needed to be made in the auditwheel_emscripten, pyodide-lock repositories as well.

### DIFF
--- a/pyodide_build/cli/build.py
+++ b/pyodide_build/cli/build.py
@@ -234,6 +234,7 @@ def _extract_extras(source_location: str) -> tuple[str, list[str]]:
 
 
 @click.command(
+    "build",
     context_settings={
         "ignore_unknown_options": True,
         "allow_extra_args": True,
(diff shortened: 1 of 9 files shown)


hoodmane (approved):
Thanks. We'll also need to update the pyodide-build submodule in Pyodide and then backport to stable branch to update the rendered docs.

ryanking13:
Good catch. Thanks. Maybe we should migrate that CLI command docs to pyodide-build doc

grden:
Thank you for the reviews. I'll create a bump PR in pyodide.

PR #427, opened 2026-08-14
install-emscripten: hint `emsdk_env.fish` for fish shell

Andrej730:
Small change to print
```
Use `source /.../0.29.4/emsdk/emsdk_env.fish` to set up the environment.
```

instead of
```
Use `source /.../0.29.4/emsdk/emsdk_env.sh` to set up the environment.
```

Kept stumbling upon trying to copy the string to source the environment and running it without modifying extension.

Relying on `SHELL` env variable is not ideal (e.g. if someone logins with `fish` and then enters `bash`, `SHELL` is still fish), but it should do for the most cases.

Tests also don't hardcode path to script anymore, which also resolves the mismatch if those tests will be ran on Windows.

Unsure if those tests will actually are unblocked on Windows, let's test in CI.

### DIFF
--- a/pyodide_build/cli/xbuildenv.py
+++ b/pyodide_build/cli/xbuildenv.py
@@ -3,7 +3,7 @@ from pathlib import Path
 import click
 
 from pyodide_build.build_env import get_build_flag, local_versions
-from pyodide_build.common import IS_WIN, default_xbuildenv_path
+from pyodide_build.common import default_xbuildenv_path, emsdk_activation_command
 from pyodide_build.views import MetadataView, SourceType
 from pyodide_build.xbuildenv import CrossBuildEnvManager
 from pyodide_build.xbuildenv_releases import (
@@ -367,8 +367,5 @@ def _install_emscripten(
     emsdk_dir = manager.install_emscripten(version, force=force)
 
     print("Installing emsdk complete.")
-    if IS_WIN:
-        cmd = f"{emsdk_dir}/emsdk_env.bat"
-    else:
-        cmd = f"source {emsdk_dir}/emsdk_env.sh"
+    cmd = emsdk_activation_command(emsdk_dir)
     print(f"Use `{cmd}` to set up the environment.")
--- a/pyodide_build/common.py
+++ b/pyodide_build/common.py
@@ -672,3 +672,25 @@ def remove_readonly(func, path, _):
     """
     os.chmod(path, stat.S_IWRITE)
     func(path)
+
+
+def emsdk_activation_command(emsdk_dir: Path) -> str:
+    """
+    Get the command to activate the given emsdk directory for the current shell.
+
+    Parameters
+    ----------
+    emsdk_dir
+        Path to the emsdk directory containing the emsdk_env script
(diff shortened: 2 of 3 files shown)


Andrej730:
```
pyodide_build/tests/test_cli_install_emscripten.py::test_install_emscripten_no_xbuildenv PASSED [  9%]
pyodide_build/tests/test_cli_install_emscripten.py::test_install_emscripten_default_version PASSED [ 18%]
pyodide_build/tests/test_cli_install_emscripten.py::test_install_emscripten_specific_version PASSED [ 27%]
pyodide_build/tests/test_cli_install_emscripten.py::test_install_emscripten_with_existing_emsdk PASSED [ 36%]
pyodide_build/tests/test_cli_install_emscripten.py::test_install_emscripten_git_failure PASSED [ 45%]
pyodide_build/tests/test_cli_install_emscripten.py::test_install_emscripten_emsdk_install_failure PASSED [ 54%]
pyodide_build/tests/test_cli_install_emscripten.py::test_install_emscripten_force_flag PASSED [ 63%]
pyodide_build/tests/test_cli_install_emscripten.py::test_install_emscripten_output_format PASSED [ 72%]
pyodide_build/tests/test_common.py::test_default_xbuildenv_path_env_var_non_writable PASSED [ 81%]
pyodide_build/tests/test_venv.py::test_windows_host_pip_wrapper_runs_with_spaces_in_paths PASSED [ 90%]
pyodide_build/tests/test_venv.py::test_windows_pyodide_cli_script_runs_with_spaces_in_paths PASSED [100%]
```

I guess those tests are unblocked now on Windows.

agriyakhetarpal:
> Relying on `SHELL` env variable is not ideal (e.g. if someone logins with `fish` and then enters `bash`, `SHELL` is still fish), but it should do for the most cases.

Do you think introducing `shellingham` or a similar dependency would be a good way to resolve this in a way that works for the vast majority of shells? If yes, I would be open to it. (If I remember correctly, it was already being picked up as a transitive dependency in our tree at some point)

Andrej730:
> Do you think introducing `shellingham` or a similar dependency would be a good way to resolve this in a way that works for the vast majority of shells? If yes, I would be open to it. (If I remember correctly, it was already being picked up as a transitive dependency in our tree at some point)

It could resolve more cases, but it seems this is such a minor issue and it's trivially fixable in case someone actually needs the fix for some shell - so maybe it's worth keeping it very simple?

agriyakhetarpal:
I think so, but I believe expanding our list of supported shells in this manner sets a precedent that can lead to the inclusion of more shells in the future, which justifies the need for `shellingham` at some point. To that effect, doing so now is better than later.

Andrej730:
If more advanced handling is needed, could it be addressed separately? I’m not sure I’d want to look into different shells detection details for this change.

ryanking13:
Maybe we can introduce `shellingham` when there is another request to support other shells? I don't really think there are any `csh` users who want to build packages, and PowerShell users can already use batch files (unless they are running powershell inside linux)

PR #429, opened 2026-08-27
use wheel_platform for pyodide-lock.json info.platform

bollwyvl:
## references
- https://github.com/pyodide/pyodide-lock/issues/44

## changes
- [x] use the [PEP-783](https://peps.python.org/pep-0783/) `pyemscripten_{:abi-version}_wasm32` platform instead of the legacy `emscripten_{:emscripten-version}_wasm32` when building new `pyodide-lock.json` files

## future work
- somewhere in the test suite, verify a built distribution's wheels and lockfile with `pyodide-lock`
  - could be on this PR


hoodmane (approved):
Makes sense to me, thanks.

